PR #12, opened 2017-10-04
Replace unwrap()s by '?'

lexxvir:
These unwrap()s lead to panic on the
real linux machine, but I don't know
what is cause of dbus error.


hwchen:
published to crates.io as 0.5.1

lexxvir:
Thanks!

hwchen:
np, thanks for the pr

PR #10, opened 2017-03-17
Add Mac keychain support

steveatinfincia:
I've added support for using the Mac keychain directly, using the security-framework crate.

**Note**: this PR should not be merged until the upstream security-framework crate merges a [PR to add password support](https://github.com/sfackler/rust-security-framework/pull/17). 

For testing purposes I have pointed this PR to a fork of security-framework, but once they merge that password support PR, I'll push another commit here to switch the cargo config back to using the upstream crate directly.

When running the included tests, be sure to disable parallel tests like this:

```
RUST_TEST_THREADS=1 cargo test
```

If the tests are run in parallel, the assert_eq!() call in the 2 round trip tests will usually fail, as the other tests might have changed the password already.

PR #26, opened 2019-04-18
Fix merge conflicts for #10

jasikpark:
I would really appreciate seeing the updated macOS keychain api merged. I hopefully fixed the merge conflicts present in #10. `RUST_TEST_THREADS=1 cargo test` succeeds as well.

This branch is also up to date with https://github.com/hwchen/keyring-rs.


hwchen:
Thanks! Appreciate you pushing this along @jasikpark . And thanks to @steveatinfincia for the original work.

hwchen:
@jasikpark I've already merged this, but wanted to note; I think that the dependency for the security-framework needs to be updated. Can you do this?

hwchen:
(From what I see, it's at `*` right now, but if you already know what minimum version it needs to be I won't need to dig around).

PR #40, opened 2020-03-24
notes on cleaning up error handling

hwchen:
Cleaning up and unifying error handling.

Currently, there are two issues with error handling in keyring:
- errors like ItemNotFound are not unified across platforms
- Error information from platform is lost when error is mapped to a more generic KeyringError.

This PR will at least make an ItemNotFound error consistent across platforms.

I hope to also reintroduce better error reporting from individual platforms, and see if other errors can be unified.

Implementation notes (on OS specific codes, etc.) are in the error module.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,3 +1,64 @@
+// TODO real error handling
+// Better unify errors so I don't have os-specific errors for important ones
+// like no password found.
+// Less important ones can get a description and put into Other
+//
+// Consider using thiserror
+//
+// ```
+// pub enum KeyringError {
+//     NoBackendFound,
+//     NoLogonSession, // windows has this, do linux and macos?
+//     NoPasswordcFound,
+//     Parse(FromUtf8Error),
+//     Other(String), // use this for things like windows ERROR_INVALID_FLAGS
+// }
+// ```
+//
+// ## Possible errors:
+//
+// Windows:
+// Call GetLastError in order to get error code to match.
+//
+// https://docs.microsoft.com/en-us/windows/win32/api/wincred/nf-wincred-creddeletew
+// - ERROR_NOT_FOUND
+// - ERROR_NO_SUCH_LOGON_SESSION
+// - ERROR_INVALID_FLAGS
+//
+// https://docs.microsoft.com/en-us/windows/win32/api/wincred/nf-wincred-credreadw
+// - ERROR_NOT_FOUND
+// - ERROR_NO_SUCH_LOGON_SESSION
+// - ERROR_INVALID_FLAGS
+// - invalid utf8 error
+//
+// https://docs.microsoft.com/en-us/windows/win32/api/wincred/nf-wincred-credwritew
+// - ERROR_NO_SUCH_LOGON_SESSION
+// - ERROR_INVALID_PARAMETER
+// - ERROR_INVALID_FLAGS
+// - ERROR_BAD_USERNAME
+// - ERROR_NOT_FOUND
+// - SCARD_E_NO_SMARTCARD
+// - SCARD_W_REMOVED_CARD
+// - SCARD_WRONG_CHV
+//
+// MacOs
+// Match on Error.code() (OSStatus)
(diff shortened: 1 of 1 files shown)


adobeDan:
I would recommend moving to `eyre` for error-handling, as its `Report` mechanism provides a consistent way to wrap lower-level errors with consistent high-level errors.  This would allow keying clients to get the lower-level information if they wish.

adobeDan:
So the way eyre works is that it has an error type called Report (which is an error in every way) that wraps other errors.  So instead of having an enum for all the different types of low-level errors, you would simply have a Report that encapsulates the lower-level errors.  Thus instead of having `enum KeyringError` you would have `type KeyringError = Report` and wherever you surfaced a lower level error you would do `.wrapError("Error accessing Keyring")` and eyre would turn that into a Report that both has a nested display capability (where the lower level message is displayed for more detail after the context message) and allows developers to retrieve the lower level error from the Report.  The eyre package itself is very small and I have found from experience that it drastically simplifies both the signatures of your public functions (they all return `eyre::Result<T>` and the implementation of them.  Take a look, for example, at the error handling in [this project](https://github.com/adobe/adobe-license-decoder.rs).

hwchen:
I poked around a bit, and it looks like `eyre` is roughly equivalent to `anyhow`. The `eyre` docs suggest using it in applications, where you mostly need to just collect and report, but use something like `thiserror` (or I guess just manually write errors, since `thiserror` just reduces boilerplate) for libraries. https://docs.rs/eyre/0.6.5/eyre/#comparison-to-thiserror

adobeDan:
Right,` eyre` is explicitly just a fork of `anyhow`, and I like its customizable error handling (which anyhow doesn't have) and the way it handles options differently.  I would have no issue with using `anyhow` if you prefer it.

The point of using something like `eyre` or `anyhow` rather than `thiserror` is exactly that it doesn't require building custom error types that distinguish the different low-level errors.  As far as I can tell any client of keyring is not going to be able to recover from the lower-level error without user intervention, so they don't actually need to distinguish what that error was programmatically, rather they need to tell the user what it was.  `eyre` and `anyhow` are perfect for that and save all the various error enums in this code.  But if you want to go with the enums then I am happy to use the `thiserror` approach.

hwchen:
Sorry, for now I'm not comfortable using `eyre`/`anyhow` in library code. It would also expose the error handling crate's error type into keyring's API, which I'd like to avoid.

I think that for now I'd like to keep the enum approach; I'm happy to consider using `thiserror` if it eases contribution, although I also want to measure that against my preference to keep dependencies to a minimum (for both maintenance and compile time reasons).

And thanks again for your discussion.

adobeDan:
I don't think it's any more complicated to keep the current approach, just more idiosyncratic.  I'm fine with doing whatever you want here.

adobeDan:
@hwchen Re-reading your comments I think now get where you're trying to go with this: you are trying to return consistent "error codes" across platforms, but in the idiomatic way of having different sub-types of the generic error type.  I'm happy to take a crack at this if you want me to.

hwchen:
I’m not quite sure what you mean, but I’m open to improvements in the error api. Can you show some small code samples of the error code and client usage?

adobeDan:
Hi @hwchen, here's the idea:

* We change the error type to be a struct rather than an enum.
* The struct has two fields:
  1. An enum with our analysis of the platform error code (one of: ItemNotFound, KeyringLocked, PlatformError, maybe one or two others that we come up with over time)
  2. The underlying platform-specific error that we interpreted and re-threw.
* Everywhere we make a platform call, we catch the error and analyze it to fill the enum and return our error struct instead (containing the underlying error).

I'm busy right now trying to rework the CI/CD scripts to use GitHub.  I can send you proposed code right after that.

hwchen:
Thanks for your work. Switching to a struct is probably a good direction, will be interested to see the code.

phillipCouto:
@hwchen is this still a valid PR to review or do we close it and wait for @adobeDan enhancement with structs?

hwchen:
Let's just close this, thanks.

PR #50, opened 2020-11-05
Remove usage of deprecated Error API (description and cause).

gamepoet:
Both description() and cause() on std::error::Error have been deprecated for a bit now (since rust 1.42 and 1.33 respectively). This patch removes description() to simply rely on the Display implementation and simply changes the cause() implementation to be source().

Since the Display and description() implementations were different, I wasn't sure which version you'd prefer to keep. I started with the just removing description() to keep the number of diffs down and if you decided you wanted this PR I can update as you so desire. :)

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -40,21 +40,7 @@ impl fmt::Display for KeyringError {
 }
 
 impl error::Error for KeyringError {
-    fn description(&self) -> &str {
-        match *self {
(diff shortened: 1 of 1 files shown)


brotskydotcom:
These changes are subsumed by #70.  Thanks for the submission; we completely agree with the update.

Issue #49, opened 2020-11-03
Can't tell if a secret exists on Windows


edmellum:
It looks to me like there's no way to get a `NoPasswordFound` from the Windows implementation which makes it hard to figure out if the library is failing or there's just no password set.

Long term maybe the API should be changed to return an `Option<T>` instead of returning an `Err` when there's no value?

hwchen:
This is pretty important, I'll try to look into this pretty soon.

hwchen:
Sorry it's taken a little while. I'm doing a round of maintenance on this project and `secret-service`, I should be getting to this after I [port](https://github.com/hwchen/secret-service-rs/pull/25) `secret-service` from dbus to zbus.

PR #51, opened 2020-11-05
get_password() on macos can now report a NoPasswordFound error

gamepoet:
I wanted to use get_password() to determine if a password already existed in the macos keychain but this was difficult without having to write platform-specific code since all errors on macos would get swallowed up into the generic MacOsKeychainError. This patch specifically looks for the errSecItemNotFound OSStatus in the error returned from the security-framework crate and maps that into a NoPasswordFound error.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -85,4 +85,11 @@ mod tests {
 
         keyring.delete_password().unwrap();
     }
+
+    #[test]
+    fn test_error_for_no_password_found() {
+        let keyring = Keyring::new(TEST_SERVICE, TEST_USER);
+        let result = keyring.get_password();
+        assert!(matches!(result, Err(KeyringError::NoPasswordFound)));
+    }
 }
--- a/src/macos.rs
+++ b/src/macos.rs
@@ -1,4 +1,4 @@
-use crate::error::Result;
+use crate::error::{KeyringError, Result};
 use security_framework::os::macos::keychain::SecKeychain;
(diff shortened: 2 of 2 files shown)


gamepoet:
Ahh ... it looks like the test I added won't succeed on windows until #49 is resolved. :-/

hwchen:
#52 should be resolved pretty soon, and then I think we should be able to get this in.

phillipCouto:
@gamepoet would you be able to resolve the conflicts the change can be reviewed.

adobeDan:
Hi @gamepoet Can you please review #70 because that also addresses this issue and the codebase has moved on quite a bit from this PR?  Thanks!

gamepoet:
Hi! :D #70 does look like it would supersede this PR if it lands - `decode_error()` in `macos.rs` does handle mapping `errSecItemNotFound` to the new `Error::NoEntry` error which would cover my need.  I can certainly look into rebasing this PR if desired, but it sounds like that might not be terribly useful with #70 on the horizon. Just let me know! :D

brotskydotcom:
Thanks @gamepoet for taking a look!  I believe we are hoping to land #70 in the next week or so.  I'm guessing that means you don't need to do anything here.  I'll close this PR as soon as we get #70 out in some form of release.

brotskydotcom:
#70 is now ready to merge, and supersedes this, so I am closing this PR.

PR #52, opened 2020-11-26
Added use of KeyringError::NoPasswordFound and KeyringError:NoBackendFound to windows code

### LINKED ISSUE #49: Can't tell if a secret exists on Windows

phillipCouto:
This resolves #49 by allowing calling code to determine if the password was not found when getting or deleting the password.

### DIFF
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -7,6 +7,8 @@ use std::os::windows::ffi::OsStrExt;
 use std::slice;
 use std::str;
 use winapi::shared::minwindef::FILETIME;
+use winapi::shared::winerror::{ERROR_NOT_FOUND, ERROR_NO_SUCH_LOGON_SESSION};
+use winapi::um::errhandlingapi::GetLastError;
 use winapi::um::wincred::{
     CredDeleteW, CredFree, CredReadW, CredWriteW, CREDENTIALW, CRED_PERSIST_ENTERPRISE,
     CRED_TYPE_GENERIC, PCREDENTIALW, PCREDENTIAL_ATTRIBUTEW,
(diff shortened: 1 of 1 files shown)


phillipCouto:
I am more than happy to flush out the error codes for the windows side but I figured I would just tackle these simple cases.

hwchen:
Thanks very much for doing this. i'll take a look asap

phillipCouto:
> Thanks very much for doing this. i'll take a look asap

No problem, more than happy to contribute 😊

hwchen:
Thanks again for doing this, handling of nopasswordfound and nobackendfound is very helpful.

For the utf16 error, it would unfortunately be a breaking change right now (because the `Error` enum is exhaustive). Can you move the utf16 code to another PR? Otherwise the rest of the code is good to go.

(Also, I might want to see if the utf16 error can be rolled up with the utf8 error into what's currently called a `Parse` error).

phillipCouto:
> Thanks again for doing this, handling of nopasswordfound and nobackendfound is very helpful.
> 
> For the utf16 error, it would unfortunately be a breaking change right now (because the `Error` enum is exhaustive). Can you move the utf16 code to another PR? Otherwise the rest of the code is good to go.
> 
> (Also, I might want to see if the utf16 error can be rolled up with the utf8 error into what's currently called a `Parse` error).

Done 😄 

Yeah I will make a new MR for handling the UTF16 error. I will try my hand at a single `Parse` for encoding errors.

phillipCouto:
@hwchen just wanted to ping and see if you have any idea when you will be able to review this.

hwchen:
Sorry, this looks good, thanks!

PR #53, opened 2020-12-14
Added proper handling of UTF-16 errors when parsing

phillipCouto:
Created a ParseError enum to serve as the new home for parse errors. Added UTF-16 to the enum and updated the windows code to now use the new `ParseError::FromUtf16` error.

Also added some comments, they may seem self explanatory but figured it is an easy quality of life add. I can remove if you want 😄

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -2,9 +2,9 @@
 use secret_service::Error as SsError;
 #[cfg(target_os = "macos")]
 use security_framework::base::Error as SfError;
-use std::error;
 use std::fmt;
 use std::string::FromUtf8Error;
+use std::{error, string::FromUtf16Error};
 
 pub type Result<T> = ::std::result::Result<T, KeyringError>;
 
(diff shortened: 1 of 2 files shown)


phillipCouto:
@hwchen and @adobeDan this is an example of a breaking change just because we added to the enum. It would be great to know how we will deal with these as it would be best if we didn't have to hold changes back until we were ready to cut a new major release as that can get messy with conflicts if the activity of the project grows.

hwchen:
For this change, I'd like to see how @adobeDan 's changes to error handling shake out first, this should be pretty easy to incorporate afterwards.

Using a struct for errors can reduce breaking changes (if I'm understanding correctly). Also, for just adding variants we can use something like `#[non_exhaustive]` which will prevent breakage for pattern matching.

adobeDan:
@phillipCouto Can you please do a thorough review of #67 and #70 and let me know if you like this direction?  If so I think we should do a major release as they will take care a lot of accumulated backlog and that way we will warn clients they need to make changes in order to adopt.

brotskydotcom:
This PR's changes are now subsumed in #70, so I'm closing this PR.

PR #55, opened 2021-01-20
Update dependencies

Victor-Savu:
In particular, updating the dependency on secret-service-rs means that
keyring-rs no longer depends on the system dbus library on linux.

I know secret-service-rs is still in pre-release, but I thought I'd open the PR just so that it is ready when the time comes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,10 +14,10 @@ default = []
 macos-specify-keychain = []
 
 [target.'cfg(target_os = "macos")'.dependencies]
-security-framework = "0.4.2"
+security-framework = "2.0.0"
 
 [target.'cfg(target_os = "linux")'.dependencies]
-secret-service = "1.1.1"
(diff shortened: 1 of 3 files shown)


hwchen:
Hey, thanks for this. Sorry I’ve been a little slow on getting key ring updated, work has been pretty busy. I’ll try to get things rolling again.

Victor-Savu:
@hwchen Thank you so much for your work on keyring-rs and on the underlying platform-specific integrations for Windows, Linux and MacOS!
I initially did the update only to serve myself because I was just eager to get rid of the native dbus dependency, but when I saw how small the change was, I decided to open a PR and share back :slightly_smiling_face:

adobeDan:
Hi @Victor-Savu, thanks so much for leading the way here!  We have now integrated the dependency updates in #66 so I closing this PR.

PR #56, opened 2021-03-16
Change Keyring construction to make it constant

jaysonsantos:
This allows code doing things like:
```rust
const KEYRING: Keyring = Keyring::new(service, username);
```

### DIFF
--- a/src/linux.rs
+++ b/src/linux.rs
@@ -2,20 +2,14 @@ use crate::error::{KeyringError, Result};
 use secret_service::{EncryptionType, SecretService};
 
 pub struct Keyring<'a> {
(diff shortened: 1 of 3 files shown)


brotskydotcom:
@jaysonsantos Why do you want to have an entry be a constant?

brotskydotcom:
With the revamp of #70 that we're about to release, it's unlikely that keyring can be constants, so I'm going to close this PR.  @jaysonsantos if you feel this is important functionality, can you please [submit an issue](https://github.com/hwchen/keyring-rs/issues/new/choose) giving the rationale for why it's needed?  Thanks!

PR #58, opened 2021-03-16
Change windows implementation to use windows-rs and change error to this error

jaysonsantos:
This is still a WIP, I  am submitting so I can see the errors on the other OSes as hopping between computers is a bit annoying.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,12 +21,18 @@ secret-service = "1.1.1"
 
 [target.'cfg(target_os = "windows")'.dependencies]
 byteorder = "1.2.1"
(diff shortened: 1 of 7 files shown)


fkrull:
I don't think switching to windows-rs is a good idea since it doesn't support crosscompiling yet (as far as I can tell, see https://github.com/microsoft/windows-rs/issues/638). Sticking with winapi until that changes seems like a safer bet to me.

brotskydotcom:
This PR's functionality is subsumed by #70, so I'm closing it.  We will merge #70 in a few days.

PR #59, opened 2021-03-23
Update secret service

jaysonsantos:
This enables services that depend on errors being Send + Sync and secret-service moved to zbus which supports it.
(Newer dbus versions also implement it)

This is dependant on https://github.com/hwchen/keyring-rs/pull/58 as it already changes the errors to thiserror


hwchen:
Thanks for this! I’ve been busy but I’ll take a look asap

brotskydotcom:
Thanks for this PR.  We have integrated the same change into #70, which we are about to merge, so I'm closing this PR.

Issue #60, opened 2021-03-28
Incompatible with 3rd party (native) credential writers who use their own Target and User Name conventions


adobeDan:
At lines 97 and 98 of windows.rs (release 0.10.1), the `TargetName` is set to be `.`-join of user name and service name.  But this is an arbitrary convention of this rust project; there is no such requirement on TargetName when used by 3rd-party credential writers.  This makes it impossible to read 3rd-party credentials written by apps that don't use a '.' somewhere in their target name.

On the other hand, I notice that the `username` field on the Rust side *is*, in fact, placed in the `UserName` field of the credential that is written.  So it's not clear to me why the `.`-join convention is used at all for the target name on the Windows side, especially since I notice that the `service` argument is used for the target name without modification in the Mac implementation.

adobeDan:
I saw the note in the sources about issue jaraco/keyring#47 and I have no problem with prepending the username to the service when setting passwords, because each writer can have their own conventions.  But then there needs to be a way of specifying the target directly on reads.  One approach would be to have a different API (e.g., one with no username parameter, where the target is specified directly).  Another might be to simply not do the concatenation if the username is empty.  I will submit a pull for the former approach, since it's more backward compatible.

adobeDan:
I'm submitting a very small PR which adopts the first option (add a windows-specific `get_password_for_target` entry which allows specifying the target explicitly).

adobeDan:
Hi @hwchen what are your feelings about this PR?  I'd like to get it merged, since it's fully backwards-compatible and is required for one client.

adobeDan:
@1Dragoon notes in #64 that we should be able to set and delete as well as read password with alternate target conventions. This seems a good point to me.

adobeDan:
Thinking about this windows-only issue, the problem with adding windows-specific `with_target` APIs is that it forces callers to write platform specific code. So perhaps these calls should be added on all platforms (although they would not have a different function there).

PR #61, opened 2021-03-29
Allow explicit target for get_password

adobeDan:
This is a windows-only change, because it's a windows-only problem.  We add a second form of get_password called `get_password_for_target`.  This fixes issue #60.

This change is fully backward-compatible in the semver sense, so it can be allowed as a minor version change (new functionality).

### DIFF
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -91,13 +91,17 @@ impl<'a> Keyring<'a> {
     }
 
     pub fn get_password(&self) -> Result<String> {
+        let target_name: String = [self.username, self.service].join(".");
+        self.get_password_for_target(&target_name)
+    }
+
+    pub fn get_password_for_target(&self, target_name: &str) -> Result<String> {
         // passing uninitialized pcredential.
         // Should be ok; it's freed by a windows api
         // call CredFree.
         let mut pcredential = MaybeUninit::uninit();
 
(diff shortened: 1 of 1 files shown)


adobeDan:
I didn't add any tests, because `get_password` now implicitly uses `get_password_for_target` so all the `get_password` apply to the new entry as well.  Since there's no way to set a password for a non-standard target there's no way to test the new entry without writing a new setter as well.  I'm happy to add the new setter if you want me to.

1Dragoon:
Could you perhaps do this so that it's also possible to get the username of the target as well? Perhaps have a function that returns a 'Credential' struct that has username and password fields, or something similar, so that both can be retrieved in one action. Would also be nice if we could store credentials without the . notation on windows targets. I'd submit a PR but I honestly don't know how this would impact Mac targets as I do not own one to test with. That, and I'm not a very experienced developer, so my code would probably suck :)

adobeDan:
Per my comments in #60, I am going to update this PR to have a setter and a deleter. I'm also going to make the new APIs work in all platforms.

hwchen:
Sounds good to me. Appreciate you moving forward on this, just give me a ping when you've added the api to the other platforms.

adobeDan:
I've decided to build this on top of the proposed update of dependencies and build systems in PR #66.  So I'm closing this PR in favor of #67.

PR #64, opened 2021-07-23
Add credential get and set methods

1Dragoon:
I'm somewhat of a noob so I feel free to criticize if this is the wrong approach. Basically I needed to be able to get username/password pairs from existing ones set by other applications and be able to set ones that other applications can use without using the . separator. There is probably a better way to do this, but I just wanted to get discussion going.

This PR shouldn't be merged as-is, as it would make windows behave differently from the other platforms, though I'm not sure of the best way to do this as I don't own a macintosh so I don't know what caveats need to be dealt with on that platform. Plus I did a few things obviously wrong, like having code duplication where it isn't necessary, though I didn't want to spend too much time as I'm 99% sure somebody will have a better approach that would render all of that work useless.

Anyways, let me know what you think :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ secret-service = "1.1.1"
 
 [target.'cfg(target_os = "windows")'.dependencies]
 byteorder = "1.2.1"
-winapi = { version =  "0.3", features = ["wincred", "minwindef"] }
+winapi = { version =  "0.3", features = ["wincred", "minwindef", "winerror"] }
 
 [dev-dependencies]
 clap = "2.0.5"
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -28,6 +28,11 @@ pub struct Keyring<'a> {
     username: &'a str,
 }
 
+pub struct TargetCredential {
+    pub username: String,
+    pub password: String,
+}
+
 impl<'a> Keyring<'a> {
     pub fn new(service: &'a str, username: &'a str) -> Keyring<'a> {
         Keyring { service, username }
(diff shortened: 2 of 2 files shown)


adobeDan:
@1Dragoon We really appreciate your starting to contribute to open source! Here are a few tips:

1. Before you contribute a PR, you should first open an issue that describes clearly what you are trying to do and what you have tried that convinced you there was a missing feature/bug.
2. Before you contribute an issue, you first search the open issues to see if there already is one that's similar that you could extend to handle your case.

In this particular case, I think you would have found issue #60 which already has a pending fix in PR #61.  You could then have tacked onto that issue that we might also need to delete entries in this case, and perhaps suggested an additional commit for the existing PR.

As to the content of your PR, it seems to address two different issues: the functionality, and a change in the windows error API.  It's not clear why you are addressing both with the same PR, so that would be a good thing to document.  As to the functional fix, introducing a special structure to hold credential info that exists only in the windows implementation is probably either too large a change (see the approach in PR #61) or one that should be propagated across all platforms (not clearly justified).

I suggest you take a look at issue #60 and comment on it if you feel delete is important functionality.  If so, I'm inclined to try to extend PR #61 to get the delete functionality rather than taking this PR forward.  Let me know what you think.

1Dragoon:
Oh yeah sorry I was aware of issue #60. I needed to address these issues for my own issues on a project I was working on, so I did the fork and then implemented the needed changes, and my project just references my own github in the cargo.toml rather than the Keyring crate proper. I mainly submitted the PR just to offer my own take on it. I didn't know you had to submit an issue first or respond to an issue. I'll go ahead and close this.

PR #68, opened 2021-11-11
Chore/update v0.10.2

hwchen:


hwchen:
@adobeDan looks like publishing the linux executable failed?

adobeDan:
Yep, there's a bug in the publish script - it shouldn't include `--locked` because we don't put the Cargo.lock file into repo.  I will fix that script and do a re-publish manually.

adobeDan:
@hwchen I have updated the script on master so it works.  I suggest you update Cargo.toml to version v0.10.4 and add release notes and let the auto-publish work correctly.  If you have already released 0.10.3 to crates.io and don't want to release a 0.10.4, then just add release notes to the 0.10.3 tag and release it and I will manually post the binaries.

adobeDan:
@hwchen one more thing: I will be retiring in a few months and would like to continue supporting this project from my non-work GitHub self @brotskydotcom .  Can you please add @brotskydotcom as a maintainer so I can work directly from that GitHub account?  Thanks!

hwchen:
I'll just release 0.10.4

Issue #65, opened 2021-08-15
KeyringError::NoPasswordFound is not used on MacOS


Macil:
On MacOS, if I try to get a nonexistent password, instead of getting an instance of KeyringError::NoPasswordFound I get `Mac Os Keychain Error: The specified item could not be found in the keychain.`.

The macos.rs file shows no uses of NoPasswordFound.

adobeDan:
This is being addressed by #70.  Stay tuned!

Issue #54, opened 2021-01-05
No backend error for Linux?


hwchen:
See https://github.com/hwchen/secret-service-rs/issues/14

if secret service api not sound, perhaps return a new error, nobackendfound, instead of generic dbus error. Or perhaps some context would be sufficient?

adobeDan:
This is being addressed by #70.  I still need to decide on all the error codes that we will surface.

brotskydotcom:
closed with merge of #70.

Issue #44, opened 2020-06-20
Add ability to access named keychains in macOS.


nagasunilt:
Current macOS implementation uses the default keychain by default and does not support opening specific keychain. This limits our ability to open system keychain since default keychain is almost always login keychain on macOS. What is the best possible to add this support to current implementation? I made this [PR](https://github.com/hwchen/keyring-rs/pull/43) but open to any better implementation if there is one.

Ideally looking for functionality similar to https://developer.apple.com/documentation/security/1396431-seckeychainopen and https://github.com/kornelski/rust-security-framework/blob/master/security-framework/src/os/macos/keychain.rs#L36

hwchen:
Thanks for making the PR. I think that the functionality looks fine, I'm just a bit wary because I haven't created any guardrails in terms of keeping the interface consistent across each OS.

I created this when I was newer at Rust, so I didn't really think about having a trait `Keyring` that each OS would implement, with extension traits specific to each OS. I guess I'm just not sure whether that would be that helpful here, though, since the library is pretty simple.

Perhaps I'll ask in one of the rust forums. And let me know if you have any feedback on this.

hwchen:
I asked the question here: https://users.rust-lang.org/t/trait-necessary-for-cross-platform-interface/44581

hwchen:
Just want to ping some recent contributors about this. Having a Keyring trait would be a breaking change.

@MaikKlein @moritzheiber @bhkaminski

naga-cf:
Hello @hwchen any update on how you want to go with this?

nagasunilt:
@hwchen any update on how you want to move forward on this?

hwchen:
I think I will just add this functionality as you’ve implemented in your pull request, and then think about using traits for a new major version. Hopefully I’ll get to it this weekend.

nagasunilt:
gentle bump!

adobeDan:
Hi @nagasunilt , we are revamping platform support for keyring and have not lost track of this issue.  Both Mac and Linux support multiple named keychains (although Mac has now deprecated using file-based keychains so only the OS ones are available).  We will look at how to support opening other than the default keychain on all platforms.  Thanks!

brotskydotcom:
closed with merge of #70

Issue #39, opened 2020-03-24
Unreliable behavior of delete_password() across operating systems


moritzheiber:
After [your comment](https://github.com/hwchen/keyring-rs/pull/38#issuecomment-602881762) I've created a short POC which (reproducably) succeeds on Linux, but fails on macOS and Windows:

```rust
extern crate keyring;

use keyring::{Keyring, KeyringError};

fn main() -> Result<(), KeyringError> {
    let service = "ce06501c7f791b4d653effb50f234cd6fa451c734054acb9936c14e965abf37d";
    let username = "username";

    let keyring = Keyring::new(service, username);
    match keyring.delete_password() {
        Err(KeyringError::NoPasswordFound) => Ok(()),
        Err(e) => Err(e),
        _ => Ok(()),
    }
}
```
On Linux this just reaches the `Ok(())` of the `NoPasswordFound` arm, on Windows it produces a `WindowsVaultError`, on macOS it returns a general `KeyringError` with the message `Mac Os Keychain Error: The specified item could not be found in the keychain.`.

I would love for the function to return a predictable result, ideally :slightly_smiling_face: 

(And have a test case for it!)

hwchen:
Thanks for this example! I'll take a look at it and have better/consistent error handling. I think that this will also affect the errors for `.get_password`.

hwchen:
I will be handling this case for sure; but I may also decide to do some more cleaning of error handling. See #40

moritzheiber:
Here are the only available error states for Secret Service (and therefore Linux): https://specifications.freedesktop.org/secret-service/latest/ch15.html

adobeDan:
@moritzheiber Just wanted you to know that this is being addressed by #70, which introduces consistent error codes across platforms.

brotskydotcom:
closed with merge of #70.

PR #70, opened 2021-11-13
Error handling improvements

adobeDan:
Changes to fix #40 - clean up error handling.

The basic idea is that errors have both a platform-independent code (to make it easier for clients to figure out what's happening) and the underlying platform error that occurred (so the client can dig into it if desired).

Right now the error codes are being determined by where in the fetch/set process the error occurs: opening the credential store, searching for the credential, reading/writing the credential, etc.  This is marked draft because I am still adding detailed underlying error analysis of the platform errors to make sure the coding is correct.  But it works on all three platforms so I welcome review.

**NOTE**: These are based on the platform reformulation done in #67, which has not been merged and introduces some minor breaking API changes.  If you have issues with that platform reformulation, please make those suggestions in #67.  The changes here could conceivably be back-ported to the existing release with a fully backward-compatible API, but I need to discuss with the other maintainers whether that's worth it or not.


adobeDan:
I believe this PR also fixes #51

README.md
```diff
@@ -15,122 +15,51 @@ __Currently supports Linux, macOS, and Windows.__ Please file issues if you have
...
+keyring = "0.10"
 ```
 
 This will give you access to the `keyring` crate in your code. Now you can use
 the `new` function to get an instance of the `Keyring` struct. The `new`
 function expects a `service` name and an `username` with which it accesses
 the password.
 
-You can get a password from the OS keyring with the `get_password` function.
+Passwords can be added to the keyring using the `set_password` function.  Then can then be read back using the `get_password` function, and deleted using the `delete_password` method.
```
phillipCouto, line 28: **suggestion:** Change `Then can then be` to `Then can be`

phillipCouto (commented):
Overall the code looks good, one comment with the struct type for error handling is it does make error handling a little less convenient at least with my current knowledge of rust. I may not be aware of rust syntax that can make this easier but I had to change the following code in our application from:

```rust
let payload = match ring.get_password() {
  Ok(s) => s,
  Err(KeyringError::NoPasswordFound) => return Ok(None),
  Err(e) => return Err(Box::new(e)),
};
```

to:

```rust
let payload = match ring.get_password() {
  Ok(s) => s,
  Err(e) => {
    if let keyring::ErrorCode::NoEntry = e.code {
      return Ok(None);
    }
    return Err(Box::new(e));
   }
};
```

For me this isn't that much of a pain but there may be others who are checking for more of the keyring error codes that may find this a negative change.

README.md
adobeDan: ha!  good catch!  I meant to say "They can then be read back" :).  Thanks I will fix it!

adobeDan:
@phillipCouto great review, thanks.  WRT to the question about client code changing, I believe you can use a destructuring pattern to minimize your code changes like this:
```rust
let payload = match ring.get_password() {
  Ok(s) => s,
  Err(keyring::Error { code: keyring::ErrorCode::NoEntry, .. })  => return Ok(None),
  Err(e) => return Err(Box::new(e)),
};
```
I also introduced type conversions between Errors and ErrorCodes, so you can easily extract the error code using .into(), but that won't work in the patten matching environment.

adobeDan:
Hi @hwchen and @phillipCouto, I believe all that's left to do now is more tests and documentation, so please give this a thorough review, especially as to issues around naming and structuring.

One serious question I have, which was raised indirectly by Phillip's comment above, is whether we want to let the error object be just an enum of codes (called "KeyringError" in the old code, and "ErrorCode" in the new code) rather than a structure.  We can always embed the platform error in the enum codes that need them, and they can even be optional.  That would really keep the API much more similar in shape to what it used to be (although I think we should call it just "Error" not "KeyringError").

brotskydotcom:
Hi @hwchen and @phillipCouto, this PR is now complete and ready to merge.  I'm going to do some code coverage analysis on the tests while you give it a final review.

brotskydotcom:
This PR:
* fixes #28 by adding a doc test for the README code.
* fixes #60 by allowing for custom maps and fetching credential attributes.
* fixes #54 by extending error codes across platforms.
* fixes #39 by extending error codes across platforms.
* fixes #65 by extending error codes across platforms.
* fixes #44 by adding keychain names on Mac and Linux.
* fixes #24 by extending error codes across platforms.

hwchen:
I'll be taking a look at this this weekend. Just wanted to pop up and say I didn't disappear, but just had a tough week.

brotskydotcom:
> I'll be taking a look at this this weekend. Just wanted to pop up and say I didn't disappear, but just had a tough week.

No worries!  I completely understand.

Also, I realize I want to revamp the examples to show the new capabilities and error handling, so let's not merge quite yet.  I'm going to put this PR back in draft to indicate that.

hwchen:
I think it's fine to have the `Error` be an enum. Now that platform errors have been consolidated, it looks clean and an enum is probably a bit more straightforward to use.

One general comment: I'm actually not sure why I originally treated passwords as utf8 strings. Perhaps it's better to treat passwords as bytes and leave encoding to the user.

Or perhaps have a `Password` type, so users could choose `as_bytes()` or `to_string()` (and we could also impl `From` both both strings and bytes). With this approach, users wouldn't have to find `String::from_utf8` themselves, and `keyring` would also be able to remove its `BadEncoding` error variant (since we'd just defer to the regular [`Utf8Error`](https://doc.rust-lang.org/std/str/struct.Utf8Error.html) from `to_string()`).

If we want to remove `String`, we can do it in another PR before releasing. Just wanted to mention it here because I was thinking about the `BadEncoding` variant.

More breaking changes, but we're already changing `Error` so this might be worth cleaning up also.

brotskydotcom:
@hwchen Thanks for the review!

With respect to non-UTF8 passwords, my thinking is that this is an edge case we should not worry about in this release.  No one has ever filed an issue saying that they need to be using arbitrary byte strings as passwords, and I suspect no one ever will.  The modern trend is increasingly to represent binary data using a UTF8-compatible serialization format (such as base 64 or JSON), rather than to store binary data directly in the platform (which requires platform-specific implementations in order to deal with different word sizes and endian-ness on different platforms).

If we heard people asking for it, it would be trivial to add `set_password_bytes` and `get_password_bytes` calls to the interface, and they would be a backwards-compatible extension that could be done as a single-dot release (assuming we are talking about 1.x here).

It's also not the case that adding support for byte strings allows getting rid of the `BadEncoding` error, because they are needed as long as we keep the string-based calls in the interface (in order to inform the caller and also return the data).

I think a much more interesting point of real extension (maybe for a 2.x release :) would be to have a `Keyring` trait with generic set and get variants that take all sorts of types.  This would allow people to plug in their own backends for the Keyring other than the ones we support, such as are available in the [Python Keyring](https://pypi.org/project/keyring/) package.  These other back ends could then support any types they wanted, and we could extend

brotskydotcom:
I'm going to merge this, so we can have all the fixes ready.  I'll do a pre-release tag so we can test with it and @hwchen can do a 1.0 release when ready.

hwchen:
Great, thanks!

PR #75, opened 2021-11-29
Reduce package keyword count to 5.

brotskydotcom:
Apparently crates.io allows at most 5 keywords.  So we can't publish without this change.


brotskydotcom:
Since this change affects nothing but the publish process on crates.io, I'm not going to bother re-tagging the GitHub release.  Instead the published crate will have one fewer keywords than the tagged crate.

Issue #86, opened 2022-06-30
Surprising feature on MacOS: empty service/account strings are wildcards


cvermilion:
The underlying [Security Framework functions](https://developer.apple.com/documentation/security/1397301-seckeychainfindgenericpassword) take optional arguments for the service and account strings, but this bubbles up to `keyring` as plain `&str`s. Passing an empty string is then treated as a non-provided optional argument rather than an explicit empty value. So, for example:

```rust
    #[test]
    fn secret_wildcard() {
        let name = "test";
        let entry = keyring::Entry::new("my_service", name);
        entry.set_password("abc").unwrap();

        let entry = keyring::Entry::new("my_service", name);
        let pass = entry.get_password().unwrap();
        assert_eq!("abc", pass);

        // Can also access via empty string(s)!
        let entry = keyring::Entry::new("my_service", "");
        let pass = entry.get_password().unwrap();
        assert_eq!("abc", pass);

        // And can use the empty-string entry to overwrite the original!
        entry.set_password("def").unwrap();

        let entry = keyring::Entry::new("my_service", name);
        let pass = entry.get_password().unwrap();
        assert_eq!("def", pass);

        entry.delete_password().unwrap();
    }
```

I've only tested this on MacOS but I suspect the behavior is the same on iOS.

This makes some sense as a reflection of the underlying API, but just reading the docs for `keyring`, I think it's a very surprising behavior. I think this should either a) get clear documentation somewhere, b) be forbidden (just fail on empty values somewhere) or c) be reflected in the API by making the service/account params `Option`s.

brotskydotcom:
Great catch, @cvermilion, thanks!  I'm inclined to simply refuse empty user names, since that could easily be made consistent behavior across all platforms, and it would conceal this Apple-specific "wildcard" feature (which can't be exposed by keyring in a cross-platform way. because it doesn't exist on Windows).

cvermilion:
Yeah, I think that's a good plan. The "search if empty" feature sort of makes sense if you're familiar with the Apple APIs but otherwise is probably unexpected, so I don't think forbidding empty fields is a big loss. If someone wants this behavior on Mac/iOS they can use `security_framework` directly.

brotskydotcom:
[...]

PR #87, opened 2022-07-01
Prevent the use of empty user, service, or target names

brotskydotcom:
This PR fixes hwchen/keyring-rs#86.

Since we can't alter the API to prevent the creation of credentials that have empty strings for target, service, or user names, we instead create "always invalid" credentials that, while well-formed, do no support calling get_password or set_password successfully.  (They always return a `NoEntry` error.)

The docs (both top-level and code-level) have been updated, and tests added to verify this handling.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ path = "examples/ios.rs"
 crate-type = ["staticlib"]
 
 [dev-dependencies]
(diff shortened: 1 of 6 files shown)


brotskydotcom:
Unless someone objects, I'm going to merge this weekend and release a new crate. @hwchen any objection?

PR #88, opened 2022-07-03
Update secret-service-rs to 2.0.2

ciehanski:
This PR updates the [secret-service-rs](https://github.com/hwchen/secret-service-rs) crate from 2.0.1 to 2.0.2. It looks like the new version was released 5 days ago.

I believe this should fix this dependabot alert I am getting on one of my [repos](https://github.com/ciehanski/kip/security/dependabot/21).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ security-framework = "2.6.1"
 security-framework = "2.6.1"
 
 [target.'cfg(target_os = "linux")'.dependencies]
-secret-service = "2.0.1"
+secret-service = "2.0.2"
 
 [target.'cfg(target_os = "windows")'.dependencies]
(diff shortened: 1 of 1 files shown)


brotskydotcom:
Hi @ciehanski, thanks for this PR.  The release yesterday of v1.2 of this repo/crate included an update of all dependencies, and that update already picked up the new secret-service.  That's why this PR doesn't include a change in Cargo.lock corresponding to the change you made in Cargo.toml.

Rather than take this PR, I'm going to start dependabot updates of all dependencies in this repo, and I'm removing all the specific (and possibly misleading) patch versions in Cargo.toml.  Hopefully that will help us keep up with any dependency changes.

Let me know if you think this PR is still needed for any reason.

PR #89, opened 2022-07-07
Add automated checks for dependency updates

brotskydotcom:
Take advantage of GitHub's dependabot to do automated checks for dependency updates.  @hwchen it would be great if you would also turn on the security update checks in the repo settings.


brotskydotcom:
Yes, that's perfect, thanks!

PR #102, opened 2022-10-29
KeyUtils Password Persistence Improvements

landhb:
This PR directly links the user's key to the session keyring to ensure that the password still exists if the Persistent keyring expires before the Session ends.

Currently the key is only linked to the persistent keyring, which is linked to the session. So the tree looks like:

```
Session Keyring
  80864054 --alswrv   1000  2002  keyring: _ses
  21973735 --alswrv   1000 65534   \_ keyring: _uid.1000
   1234379 ---lswrv   1000 65534   \_ keyring: _persistent.1000
 641572533 --alswrv   1000  2002       \_ user: keyring-rs:landhb@vault
```

So when the `_persistent` keyring expires, so will our key. Since after expiration the tree will look like:

```
Session Keyring
  80864054 --alswrv   1000  2002  keyring: _ses
  21973735 --alswrv   1000 65534   \_ keyring: _uid.1000
```

But if we link to both the persistent keyring and the session keyring, our key is still around if the persistent keyring expires before the session ends.

```
Session Keyring
  80864054 --alswrv   1000  2002  keyring: _ses
  21973735 --alswrv   1000 65534   \_ keyring: _uid.1000
 238306210 --alswrv   1000  2002   \_ user: keyring-rs:landhb@vault
   1234379 ---lswrv   1000 65534   \_ keyring: _persistent.1000
 238306210 --alswrv   1000  2002       \_ user: keyring-rs:landhb@vault
```

After expiration the key will still be available now:

```
Session Keyring
  80864054 --alswrv   1000  2002  keyring: _ses
  21973735 --alswrv   1000 65534   \_ keyring: _uid.1000
 238306210 --alswrv   1000  2002   \_ user: keyring-rs:landhb@vault
```

And next time they call `get_password()` it will re-link it to the persistent keyring, allowing the password to persist across a logout/login again.

This aligns more closely with what users of `keyring-rs` expect. Since the password will still be available as long as either the persistent keyring is valid OR the session is valid.

| Persistent Keyring State | Session Keyring State | User Key State |
| -------------            | -------------         | -------------  | 
| Active                   | Active                | Active         |
| Expired                  | Active                | Active         |
| Active                   | Logged Out            | Active (Accessible on next login)        |
| Expired                  | Logged Out            | Expired        |

### DIFF
--- a/src/keyutils.rs
+++ b/src/keyutils.rs
@@ -15,8 +15,10 @@ use super::error::{decode_password, Error as ErrorCode, Result};
 /// set_password is called.
 #[derive(Debug, Clone)]
 pub struct KeyutilsCredential {
-    /// Host keyring
-    pub inner: KeyRing,
+    /// Host session keyring
+    pub session: KeyRing,
+    /// Host persistent keyring
+    pub persistent: KeyRing,
     /// Description of the key entry
     pub description: String,
 }
@@ -36,14 +38,12 @@ impl CredentialApi for KeyutilsCredential {
 
         // Add to the persistent keyring
         let key = self
-            .inner
+            .session
             .add_key(&self.description, password)
             .map_err(decode_error)?;
 
-        // Directly link to the Session keyring as well
-        let session =
-            KeyRing::from_special_id(KeyRingIdentifier::Session, false).map_err(decode_error)?;
-        session.link_key(key).map_err(decode_error)?;
+        // Directly link to the persistent keyring as well
+        self.persistent.link_key(key).map_err(decode_error)?;
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


brotskydotcom (changes requested):
Hi @landhb, sorry it's taken me so long to review this.

I love these changes; thanks so much!

The only issue I have is that this crate requires the use of `rustfmt` because, while I don't always love what it does, it removes religion from the code reviews :).  And I think some of your (excellent, don't get me wrong) formatting changes will not be acceptable to `rustfmt` (e.g., your change in order of `use` statements, or some of your "unnecessary" line breaking).

Can you please do me the favor of running `rustfmt` on this file and submitting the result?  At that point I will be happy to issue an alpha.2 with these changes.  (I'm still in alpha because I haven't had time to figure out what I'm doing with secret service yet :)

brotskydotcom (approved):
Great, thanks for being careful!

PR #107, opened 2023-01-16
Fix #85 by providing useful error on oversized pw

### LINKED ISSUE #85: Error code 1783 on Windows
When trying to add a keyring entry on Windows, I get `Error code 1783` which seems to correlate to `RPC_X_BAD_STUB_DATA` in winapi.

Alexei-Barnes:
Left a limerick in the password in the test I added for good measure.

### DIFF
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -131,7 +131,9 @@ impl WinCredential {
                 CRED_MAX_STRING_LENGTH,
             ));
         }
-        if password.len() > CRED_MAX_CREDENTIAL_BLOB_SIZE as usize {
+        // We're going to store the password as UTF-16, so make sure to consider its length as UTF-16.
+        // `encode_utf16` gives us the count of `u16`s, so we multiply by 2 to get the number of bytes.
+        if password.encode_utf16().count() * 2 > CRED_MAX_CREDENTIAL_BLOB_SIZE as usize {
             return Err(ErrorCode::TooLong(
                 String::from("password"),
                 CRED_MAX_CREDENTIAL_BLOB_SIZE,
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -414,7 +414,7 @@ mod tests {
             ("target", CRED_MAX_GENERIC_TARGET_NAME_LENGTH),
             ("target alias", CRED_MAX_STRING_LENGTH),
             ("comment", CRED_MAX_STRING_LENGTH),
-            ("password", CRED_MAX_CREDENTIAL_BLOB_SIZE),
+            ("password", CRED_MAX_CREDENTIAL_BLOB_SIZE / 2),
         ] {
             let long_string = generate_random_string_of_len(1 + len as usize);
             let mut bad_cred = cred.clone();
@@ -427,8 +427,30 @@ mod tests {
                 "password" => password = &long_string,
                 other => panic!("unexpected attribute: {}", other),
             }
-            validate_attribute_too_long(bad_cred.validate_attributes(password), attr, len);
+            let expected_length = if attr == "password" { len * 2 } else { len };
+            validate_attribute_too_long(
+                bad_cred.validate_attributes(password),
+                attr,
+                expected_length,
+            );
         }
+    }
+
+    #[test]
+    fn test_password_valid_only_after_conversion_to_utf16() {
+        let cred = WinCredential {
+            username: "username".to_string(),
+            target_name: "target_name".to_string(),
+            target_alias: "target_alias".to_string(),
+            comment: "comment".to_string(),
+        };
+
+        let len = CRED_MAX_CREDENTIAL_BLOB_SIZE / 2;
+        let password: String = (0..len).map(|_| "笑").collect();
+
+        assert!(password.len() > CRED_MAX_CREDENTIAL_BLOB_SIZE as usize);
+        cred.validate_attributes(&password)
+            .expect("Password of appropriate length in UTF16 was invalid");
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


Alexei-Barnes:
This fixes #85

src/windows.rs
```diff
@@ -131,7 +131,9 @@ impl WinCredential {
                 CRED_MAX_STRING_LENGTH,
             ));
         }
-        if password.len() > CRED_MAX_CREDENTIAL_BLOB_SIZE as usize {
+        // We're going to store the password as UTF-16, so make sure to consider its length as UTF-16.
+        // `encode_utf16` gives us the count of `u16`s, so we multiply by 2 to get the number of bytes.
+        if password.encode_utf16().count() * 2 > CRED_MAX_CREDENTIAL_BLOB_SIZE as usize {
```
brotskydotcom, line 136: Thanks, this is the perfect test, and a beautifully minimal change!

tests/basic.rs
```diff
@@ -77,6 +77,25 @@ fn test_round_trip_non_ascii_password() {
...
...1pIHNpdCBhbWV0IGZldWdpYXQuIE51bmMgdmVoaWN1bGEgb2RpbyBlZ2V0IHZlaGljdWxhIHBvc3VlcmUuIFNlZCBwdWx2aW5hciBkaWFtIHZpdGFlIHNhcGllbiBncmF2aWRhIG1hdHRpcy4gVml2YW11cyBzZWQgcHVydXMgc29sbGljaXR1ZGluLCBsYWNpbmlhIGVuaW0gZXQsIGRpZ25pc3NpbSBvcmNpLiBJbiBpZCBjdXJzdXMgc2FwaWVuLCBpbiBjb25ndWUgbWFzc2EuIFBoYXNlbGx1cyBzaXQgYW1ldCByaXN1cyBub24gc2VtIHZpdmVycmEgYXVjdG9yLiBTZWQgdGVsbHVzIG51bGxhLCB2ZXN0aWJ1bHVtIGNvbW1vZG8gY3Vyc3VzIGEsIGlhY3VsaXMgaW4gZXguClV0IGV4IGxpYmVybywgdnVscHV0YXRlIGFjIHR1cnBpcyBldSwgZWdlc3RhcyB2YXJpdXMgZHVpLiBJbnRlZ2VyIGlkIHF1YW0gZWxpdC4gUHJvaW4gZmF1Y2lidXMgZ3JhdmlkYSBtZXR1cywgcnV0cnVtIHJob25jdXMgbWV0dXMgY29uc2VxdWF0IHF1aXMuIE51bGxhIGlkIHRvcnRvciBzYXBpZW4uIFBlbGxlbnRlc3F1ZSBjb25ndWUgZXN0IGxhY3VzLCBpbiB0ZW1wb3Igb3JjaSBwb3N1ZXJlIHZlbC4gU2VkIGRhcGlidXMgb3JuYXJlIHRyaXN0aXF1ZS4gTWF1cmlzIGxlbyBudW5jLCB0aW5jaWR1bnQgZXQgc2VtIHNlZCwgbWFsZXN1YWRhIGhlbmRyZXJpdCBsb3JlbS4gUHJhZXNlbnQgZmF1Y2lidXMgcGxhY2VyYXQgdmVuZW5hdGlzLiBBbGlxdWFtIGVyYXQgdm9sdXRwYXQuIEN1cmFiaXR1ciBzYWdpdHRpcyBlbGVtZW50dW0gbGliZXJvLiBQcmFlc2VudCBzZWQgbWFsZXN1YWRhIGFyY3UsIHNlZCBsYW9yZWV0IGVyYXQuIEN1cmFiaXR1ciBldCBncmF2aWRhIHR1cnBpcy4gQWVuZWFuIGxhY2luaWEgZXggYXQgZXJvcyB2b2x1dHBhdCB1bGxhbWNvcnBlci4gQWVuZWFuIGF0IHNjZWxlcmlzcXVlIGFudGUu";
```
brotskydotcom, line 84: This looked like base64 so I decoded it and found your limerick (which has a typo, by the way).  Cute!

Unfortunately, this is not the right way to test your improved error check.  When you encode UTF-8 as UTF-16, the number of UTF-16 bytes will never be more than twice the number of UTF-8 bytes, so if all we're going to do is test with ascii, we could have just used the old error check and divided the allowed max length by 2.  Since UTF-16 is more compact than UTF-8 for characters in the basic multilingual plane (U+0000 to U+FFFF), the right test for your improved error check is to generate a UTF-8 string that's too long in UTF-8 but whose characters are in that plane so it shrinks to being storable in UTF-16.  That is, we want a positive test as well as a negative test.

So please do the following:
- Change this test to be a positive test that puts in a too-long UTF-8 string that is storable after conversion to UTF-16.
- Update the existing validate-attributes test so that it generates an alphanumeric string that is half the max bytes + 1 byte, because when that is expanded it should be too long.

(If you're busy, you can just pull this test completely and I will write the tests.  I really appreciate you're figuring out the minimal change needed in validate_attributes.

brotskydotcom (changes requested):
Great work, thanks.  Please see my notes about the testing.

tests/basic.rs
```diff
@@ -77,6 +77,25 @@ fn test_round_trip_non_ascii_password() {
     )
 }
 
+#[test]
```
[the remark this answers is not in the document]
Alexei-Barnes: I don't think a separate round trip test is needed to validate this since the unit test coverage is clearly sufficient.

tests/basic.rs
Alexei-Barnes: Thanks for the review - I've modified the tests as suggested.

brotskydotcom (approved):
Really good work, thanks!  I will back-integrate it into 1.2 as well.

PR #124, opened 2023-05-17
Remove unnecessary clone/allocation in `decode_password`

phlip9:
Just a small fix I noticed while integrating `keyring` : )

The error returned by `String::from_utf8` actually gives us access to the original buffer: [`FromUtf8Error::into_bytes`](https://doc.rust-lang.org/stable/std/string/struct.FromUtf8Error.html#method.into_bytes). Using this lets us avoid an extra clone on the password buffer every time.


brotskydotcom (approved):
Thanks for this!  That extra clone has always driven me crazy!

PR #152, opened 2023-12-20
docs: document platform-specific behavior on Windows 

ReactorScram:
where `service` is not used.

This could lead to callers silently using a very generic collision-prone name like "token"

e.g. https://github.com/firezone/firezone/blob/main/rust/windows-client/src-tauri/src/client/gui.rs#L215-L219

I ran `cargo test` a few times and it passed once but failed other times. I think it isn't playing well with Rust's multi-threaded test framework.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -211,6 +211,10 @@ impl Entry {
     /// Create an entry for the given target, service, and user.
     ///
     /// The default credential builder is used.
+    ///
+    /// # Platform-specific behavior
+    ///
+    /// On Windows, `service` is ignored.
     pub fn new_with_target(target: &str, service: &str, user: &str) -> Result<Entry> {
         build_default_credential(Some(target), service, user)
     }


brotskydotcom:
Hi @ReactorScram, thanks for the suggestion.  All the platform-specific behavior is documented in the platform-specific documentation, so you might want to [switch your platform on docs.rs to Windows](https://docs.rs/keyring/2.1.0/i686-pc-windows-msvc/keyring/index.html) (rather than the default linux).  There, in the documentation of the [windows credential store module](https://docs.rs/keyring/2.1.0/i686-pc-windows-msvc/keyring/windows/index.html), you will find this paragraph:

> Because the Windows credential manager doesn't support multiple collections of credentials,
and because many Windows programs use _only_ the service name as the credential _target name_,
the `Entry::new_with_target` call uses the `target` parameter as the credential's _target name_
rather than concatenating the username and service.
So if you have a custom algorithm you want to use for computing the Windows target name,
you can specify the target name directly.  (You still need to provide a service and username,
because they are used in the credential's metadata.)

With respect to the testing issue you've noticed, I have found that the problem seems to be less related to multi-threading and more to proper initialization of the windows credential manager (which can be tricky in a terminal session, especially if your are running headless).  Whenever I'm actually on a physical windows box, and have logged in "normally", the tests run fine.

I'm going to close this PR based on the above.  If you feel the current docs aren't adequate, please feel free to open an issue.

PR #160, opened 2024-01-01
Replace lazy_static with OnceLock

russellbanks:
This pull request removes [`lazy_static`](https://crates.io/crates/lazy_static) in favour of [OnceLock](https://doc.rust-lang.org/stable/std/sync/struct.OnceLock.html). [OnceCell](https://github.com/matklad/once_cell) is technically a crate. However, it has been [partially merged](https://github.com/rust-lang/rust/pull/105587) into the standard library.

lazy_static hasn't had a release in 4 years and now has a [section](https://github.com/rust-lang-nursery/lazy-static.rs#standard-library) on its ReadMe that shows how to achieve the exact same thing with the standard library.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,9 @@ linux-secret-service-rt-tokio-crypto-openssl = ["secret-service/rt-tokio-crypto-
 linux-no-secret-service = ["linux-default-keyutils"]
 linux-default-keyutils = ["linux-keyutils"]
 
+[dependencies]
+once_cell = "1"
+
 [target.'cfg(target_os = "macos")'.dependencies]
 security-framework = { version = "2.6", optional = true }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,9 +28,6 @@ linux-secret-service-rt-tokio-crypto-openssl = ["secret-service/rt-tokio-crypto-
 linux-no-secret-service = ["linux-default-keyutils"]
 linux-default-keyutils = ["linux-keyutils"]
 
-[dependencies]
-once_cell = "1"
-
 [target.'cfg(target_os = "macos")'.dependencies]
 security-framework = { version = "2.6", optional = true }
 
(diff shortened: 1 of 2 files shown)


brotskydotcom (approved):
These changes look good, but I don't want to merge this yet.  See other comments in the PR.

brotskydotcom:
I don't want to move the minimum build platform for this crate to 1.70 (which is required by these changes).  There are a lot of clients on second-tier platforms (such as the BSDs), and I notice that the built-in rust on those platforms can be a year or more old.  The crate is built for the 2021 edition (initially Rust 1.56), so I'm trying to keep that as the minimum build platform (at least for v2).

I'm happy to revisit this when the 2024 edition ships, and I expect to release a v3 somewhere around then that will integrate this change.  Marking this PR as waiting until then.

russellbanks:
Until then, we could use the actual OnceCell crate? It has an MSRV of 1.60. We'd still benefit from it being maintained and it achieves the same thing without using a macro.

russellbanks:
I've replaced this with Lazy from OnceCell which more closely matches lazy_static. Once LazyCell (standard library equivalent of Lazy) gets stabilised and/or we move to a higher MSRV, we can replace this with OnceLock or LazyCell.

brotskydotcom (approved):
These changes look fine.  Again, see comments further down.

brotskydotcom:
How about, instead of just switching crates from lazy_static to OnceCell, we instead leave lazy_static in place for older compilers, add the rustversion crate, and use the stable language feature you were originally suggesting on MSRV of 1.70 or higher?  Since almost everyone is on 1.70 anyway, I'd rather make a chance to the stable language feature now instead of just picking up a different crate dependency that affects everyone (and that i would have to carefully test on 1.68 to be sure there's no change).

brotskydotcom:
I took these changes as part of #179. Thanks so much for your contribution!

PR #161, opened 2024-01-01
Add .idea to .gitignore

russellbanks:
`.idea` is a config folder created by JetBrains IDEs, notably RustRover, and is helpful to ignore for contributing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,5 @@ libtest.rmeta
 # profiling
 *.profdata
 *.profraw
-
+# JetBrains IDEs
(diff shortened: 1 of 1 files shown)


brotskydotcom:
I also use JetBrains IDEs, but I don't particularly want an artifact in this project that is IDE-specific.  I rely on contributors to be careful about keeping their IDE artifacts out of pull requests (e.g., by using a global git ignore for them).

So, thanks much for the suggestion, and I take your point, but I'm not going to include this change.

PR #172, opened 2024-06-06
Add store password and get password blob functions for windows

zschreur:
Because of how the credentials are converted from u8 to u16 when stored on windows, it is not possible to store a blob of data that is greater than `CRED_MAX_CREDENTIAL_BLOB_SIZE / 2`.

This PR provides a workaround by creating a `set_password_blob` and `get_password_blob` function. These functions take and return a `Vec<u8>`. This allows a client to do something like:
```rs
pub fn store_cred(service: &str, user: &str, password: &Vec<u8>) -> Result<()> {
    match keyring::Entry::new(service, key)?
        .get_credential()
        .downcast_ref::<keyring::windows::WinCredential>()
    {
        Some(win_credential) => {
            win_credential.set_password_blob(password)
        },
       // ...
```

The code for getting & setting with a `Vec<u8>` was pulled out of `set_password` and `get_password`, and then those functions are now making use of the new blob functions.

### DIFF
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -57,31 +57,20 @@ pub struct WinCredential {
     pub comment: String,
 }
 
-// Windows API type mappings:
-// DWORD is u32
-// LPCWSTR is *const u16
-// BOOL is i32 (false = 0, true = 1)
-// PCREDENTIALW = *mut CREDENTIALW
-
-impl CredentialApi for WinCredential {
+impl WinCredential {
     /// Create and write a credential with password for this entry.
+    ///
+    /// Password is stored as u8 byte sequence rather than converting it to u16
     ///
     /// The new credential replaces any existing one in the store.
     /// Since there is only one credential with a given _target name_,
     /// there is no chance of ambiguity.
-    fn set_password(&self, password: &str) -> Result<()> {
-        self.validate_attributes(password)?;
+    pub fn set_password_blob(&self, blob: &Vec<u8>) -> Result<()> {
+        self.validate_attributes_blob(blob)?;
         let mut username = to_wstr(&self.username);
         let mut target_name = to_wstr(&self.target_name);
         let mut target_alias = to_wstr(&self.target_alias);
         let mut comment = to_wstr(&self.comment);
-        // Password strings are converted to UTF-16, because that's the native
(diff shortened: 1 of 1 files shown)


brotskydotcom:
Hi @zschreur, thanks for contributing this PR.  Before I can carefully consider it, I need to understand what your design goals are and what motivates them.  Can you please open a bug describing the issue you're trying to solve for? And then can you add some design notes to this PR explaining what solution you chose and why?

Here are some things to think about as you write the bug:

- Are you trying to use keyring to write arbitrary binary data into credentials?  That would seem to require changes on all platforms.
- Are you trying to write long credential data? There are limitations in the length of the credential store on most platforms; why should Windows be treated differently?

Here is something to think about as you write your design note:

- Keyring is meant to be interoperable with 3rd party use of the credential store, such as the Windows OS credential manager UI. Storing passwords as blobs would break that interoperability.

Thanks again for contributing. I look forward to having more discussion about this.

zschreur:
@brotskydotcom thank you for considering my PR. I have written up the issue with as much detail as I can here: https://github.com/hwchen/keyring-rs/issues/173.

The approach I took when designing this PR was to provide a fix for #173, while not bringing in a breaking change. This solution seemed to be the least invasive.

In my opinion I would have expected `get_password` and `set_password` to behave in the same way I have written `get_password_blob` and `set_password_blob` to work. So in that scenario, this PR does not actually solve that issue.

> Keyring is meant to be interoperable with 3rd party use of the credential store, such as the Windows OS credential manager UI. Storing passwords as blobs would break that interoperability.

In regards to interoperability, currently the way that credentials are stored on Windows the Windows OS Credential Manager UI does not allow for displaying the stored credential. You are only given the ability to delete or replace it. Because of this, I believe that storing the `&str` as its `&[u8]` byte slice representation does not break interoperability.

brotskydotcom:
Hey @zschreur, did you make any progress on reworking this PR to be generic a la #174?

brotskydotcom:
I have written this functionality into the v3 release (now at rc.2), so I'm closing this PR.

zschreur:
> I have written this functionality into the v3 release (now at rc.2), so I'm closing this PR.

Great thank you for letting me know.

PR #186, opened 2024-07-10
Release v3.0.1

brotskydotcom:
Primary change: Make errors be Send + Sync. This is what they were in v2, and it's required by anyhow.

Fixes hwchen/keyring#185.

Also updates dbus-secret-service dependency.


tmpfs:
Thanks for the quick fix @brotskydotcom and the patch release 👍

PR #188, opened 2024-07-13
Add set_secret() and get_secret() for iOS impl.

### LINKED ISSUE #187: IosCredentials is missing trait methods in v3
```
Showing All Messages
SEVERE:   --> /Users/muji/.cargo/registry/src/index.crates.io-6f17d22bba15001f/keyring-3.0.1/src/ios.rs:38:1
SEVERE:    |
SEVERE: 38 | impl CredentialApi for IosCredential {
SEVERE:    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `set_secret`, `get_secret` in implementation
SEVERE:    |
SEVERE:   ::: /Users/muji/.cargo/registry/src/index.crates.io-6f17d22bba15001f/keyring-3.0.1/src/credential.rs:24:5
SEVERE:    |
SEVERE: 24 |     fn set_secret(&self, password: &[u8]) -> Result<()>;
SEVERE:    |     ---------------------------------------------------- `set_secret` from trait
SEVERE: ...
SEVERE: 34 |     fn get_secret(&self) -> Result<Vec<u8>>;
SEVERE:    |     ---------------------------------------- `get_secret` from trait
SEVERE: 
SEVERE: For more information about this error, try `rustc --explain E0046`.
```

tmpfs:
Closes #187.

### DIFF
--- a/src/ios.rs
+++ b/src/ios.rs
@@ -47,6 +47,16 @@ impl CredentialApi for IosCredential {
         Ok(())
     }
 
+    /// Create and write a credential with secret for this entry.
+    ///
+    /// The new credential replaces any existing one in the store.
+    /// Since there is only one credential with a given _account_ and _user_
+    /// in any given keychain, there is no chance of ambiguity.
+    fn set_secret(&self, secret: &[u8]) -> Result<()> {
+        set_generic_password(&self.service, &self.account, secret).map_err(decode_error)?;
+        Ok(())
+    }
+
     /// Look up the password for this entry, if any.
     ///
     /// Returns a [NoEntry](ErrorCode::NoEntry) error if there is no
@@ -55,6 +65,14 @@ impl CredentialApi for IosCredential {
         let password_bytes =
             get_generic_password(&self.service, &self.account).map_err(decode_error)?;
         decode_password(password_bytes.to_vec())
+    }
+
+    /// Look up the secret for this entry, if any.
+    ///
+    /// Returns a [NoEntry](ErrorCode::NoEntry) error if there is no
+    /// credential in the store.
+    fn get_secret(&self) -> Result<Vec<u8>> {
+        get_generic_password(&self.service, &self.account).map_err(decode_error)
     }
 
     /// Delete the underlying generic credential for this entry, if any.
--- a/src/ios.rs
+++ b/src/ios.rs
@@ -62,7 +62,7 @@ impl CredentialApi for IosCredential {
     /// credential in the store.
     fn get_password(&self) -> Result<String> {
         let password_bytes = self.get_secret()?;
-        decode_password(password_bytes.to_vec())
+        decode_password(password_bytes)
     }
 
     /// Look up the secret for this entry, if any.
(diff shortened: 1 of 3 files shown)


brotskydotcom (changes requested):
Your definitions of `set_secret` and `get_secret` are perfect.

But, like the macOS implementation, you should have `set_password` call `set_secret` like this:
```
    fn set_password(&self, password: &str) -> Result<()> {
        self.set_secret(password.as_bytes())
    }
```
and you should have `get_password` call `get_secret` like this:
```
    fn get_password(&self) -> Result<String> {
        let password_bytes = self.get_secret()?;
        decode_password(password_bytes.to_vec())
    }
```

tmpfs:
Good call @brotskydotcom , thats done!

brotskydotcom (approved):
Excellent! Thank you so much!

tmpfs:
I notice you convert to `Vec<u8>` when calling `decode_password()` but `std::str::from_utf8` will take a slice so you can just pass to without the calls to `to_vec()`.

brotskydotcom:
Can you please update README.md to add your name in the contributor's list?

brotskydotcom:
> I notice you convert to `Vec<u8>` when calling `decode_password()` but `std::str::from_utf8` will take a slice so you can just pass to without the calls to `to_vec()`.
> 
> Edit: so if you prefer `std::str::from_utf8` to `String::from_utf8` then we could avoid the allocations here.

Well the definition of decode_password is shared across platforms and most of the others have already done the vector conversion.  But I still like the optimization you suggest in the iOS case, because password_bytes is already a Vec<u8> when it comes back from `get_secret`, so there's really no reason to have the `.to_vec()` redundant copy made.  It should just be `decode_password(password_bytes)`.

src/ios.rs
```diff
@@ -52,11 +61,18 @@ impl CredentialApi for IosCredential {
     /// Returns a [NoEntry](ErrorCode::NoEntry) error if there is no
     /// credential in the store.
     fn get_password(&self) -> Result<String> {
-        let password_bytes =
-            get_generic_password(&self.service, &self.account).map_err(decode_error)?;
+        let password_bytes = self.get_secret()?;
         decode_password(password_bytes.to_vec())
```
brotskydotcom, line 65: Please remove this call, as it's unnecessary.  `password_bytes` is already a Vec<u8>.

brotskydotcom (changes requested):
Just one last fix needed.

brotskydotcom:
Whoops.  Just noticed one more fix needed to get the tests working on iOS:

You need to change line 91 from 
```
    use keyring::default::IosCredential;
```
to
```
    use keyring::ios::IosCredential;
```

src/ios.rs
tmpfs: Agreed, thats done.

brotskydotcom (approved):
Perfect, thanks for keeping at it!

PR #193, opened 2024-07-15
Fix linux-native feature condition

Brooooooklyn:


brotskydotcom (approved):
These changes look good, thanks.  Can you please file a bug about this and explain exactly what happened that made you notice the problem?  (For example: "I turned on the linux-native feature but ended up with the mock credential store as the default.")

brotskydotcom:
Thanks for noticing this issue, Brooklyn!  It turns out there are some other code issues that your fix reveals, so I'm going to close this PR, file the bug, and the make a new release that incorporates your changes (crediting you, of course).  Hopefully the new release will be out before you even get a chance to read this comment :).

PR #210, opened 2024-09-09
Add error and panic docs to entry associated function new

ShaunSHamilton:
Adds some docs to `Entry::new` about errors and panics.

I spent too long figuring out why the return was a `Result` when the docs do not mention what could go wrong.

Feel free to adjust wording as desired.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -311,16 +311,14 @@ impl Entry {
     ///
     /// # Errors
     ///
-    /// This function will return an [`Error`] if the `service` or `user` values are invalid.
+    /// This function will return an [Error] if the `service` or `user` values are invalid.
     /// The specific reasons for invalidity are platform-dependent, but include length constraints.
-    ///
-    /// [`Error`]: Error
     ///
     /// # Panics
     ///
-    /// This function may panic if the internal credential builder's `RwLock` is poisoned. This
-    /// would typically happen if a thread panicked while holding a lock. If you encounter this,
-    /// it is considered a serious bug and should be reported.
+    /// In the very unlikely event that the internal credential builder's `RwLock`` is poisoned, this function
+    /// will panic. If you encounter this, and especially if you can reproduce it, please report a bug with the
+    /// details (and preferably a backtrace) so the developers can investigate.
     pub fn new(service: &str, user: &str) -> Result<Entry> {
         build_default_credential(None, service, user)
     }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -308,6 +308,19 @@ impl Entry {
     /// Create an entry for the given service and user.
     ///
     /// The default credential builder is used.
+    ///
+    /// # Errors
+    ///
+    /// This function will return an [`Error`] if the `service` or `user` values are invalid.
```
brotskydotcom, line 314: Please don't use backticks around links, so [Error] not [`Error`] (see examples at the top of the file).

src/lib.rs
```diff
@@ -308,6 +308,19 @@ impl Entry {
...
+    /// The specific reasons for invalidity are platform-dependent, but include length constraints.
+    ///
+    /// [`Error`]: Error
```
brotskydotcom, line 317: This reference is not necessary if you remove the backticks, as requested above.

src/lib.rs
```diff
@@ -308,6 +308,19 @@ impl Entry {
...
+    ///
+    /// # Panics
+    ///
+    /// This function may panic if the internal credential builder's `RwLock` is poisoned. This
+    /// would typically happen if a thread panicked while holding a lock. If you encounter this,
+    /// it is considered a serious bug and should be reported.
```
brotskydotcom, line 323: I'd rather not get into why an `RwLock` might be poisoned.  How about something like this instead:

In the very unlikely event that the internal credential builder's `RwLock` is poisoned, this function
will panic.  If you encounter this, and especially if you can reproduce it, please report a bug with the
details (and preferably a backtrace) so the developers can investigate.

brotskydotcom (changes requested):
Thanks much for this update; it's great to have these sections in the docs.  Can you please look at my review suggestions and update accordingly?

src/lib.rs
ShaunSHamilton: Sure. I was actually just copying the std library format: https://doc.rust-lang.org/stable/src/std/sync/rwlock.rs.html#292-301

brotskydotcom:
Just FYI, I probably won't publish a release with these changes for several weeks, because I'm working on some code changes and will wait for them to be done.  But thanks again for the contribution!

Issue #214, opened 2024-09-21
Invalid `compile_error` condition


soywod:
https://github.com/hwchen/keyring-rs/blob/27a7b3514ee7535374637955f682cf16eeaa25fa/src/lib.rs#L173-L181

I think the condition is not correct, at least for Linux. It prevents Linux users to combine both the secret service based keystore (for persistent storage, on disk) and the keyutils (for cache, in memory). I actually have a combination of both with your `v2`, and this compile error prevents me to upgrade to the `v3`:

```toml
keyring_native = { version = "3", package = "keyring", default-features = false, features = ["linux-native", "apple-native", "windows-native", "async-secret-service"] }
```

```text
error[E0252]: the name `default` is defined multiple times
   --> /home/soywod/.cargo/registry/src/index.crates.io-6f17d22bba15001f/keyring-3.3.0/src/lib.rs:208:9
    |
197 | pub use keyutils as default;
    |         ------------------- previous import of the module `default` here
...
208 | pub use secret_service as default;
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^ `default` reimported here
    |
    = note: `default` must be defined only once in the type namespace of this module
help: you can use `as` to change the binding name of the import
    |
208 | pub use secret_service as other_default;
    |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

error: You can enable at most one keystore per target architecture
   --> /home/soywod/.cargo/registry/src/index.crates.io-6f17d22bba15001f/keyring-3.3.0/src/lib.rs:181:1
    |
181 | compile_error!("You can enable at most one keystore per target architecture");
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

brotskydotcom:
I completely agree that this is a bug, Clément, and should be fixed.  But I have some question about how to fix it.

When I first did v3, I figured I would just have the two secret-service modules be conflicting, and let you combine them with the keyutils store. But that raised the question of which of the two services (secret-service or keyutils) should be the default. I really didn't like the v2 features that let you set the default, so instead I made them all conflict, figuring I would force people to choose which one they wanted.  But that eliminates the ability to combine them, as you discovered :).

[...]

PR #215, opened 2024-09-22
Allow use of both linux-native and secret-service keystores.

brotskydotcom:
Using both will make the mock keystore be the default, so clients will have to use `new_with_credential` to wrap their platform-specific credentials.

Fixes hwchen/keyring-rs#214.


src/lib.rs
```diff
@@ -203,17 +197,24 @@ pub use keyutils as default;
...
 pub use secret_service as default;
 
 #[cfg(all(
     target_os = "linux",
-    not(any(
-        feature = "linux-native",
-        feature = "sync-secret-service",
-        feature = "async-secret-service"
-    ))
+    any(
```
soywod, line 207: This condition illustrates well what I tried to explain here: https://github.com/hwchen/keyring-rs/issues/214#issuecomment-2365500645.

It just looks wrong and messy. Also not sustainable: how will it look like in few years, with other new keystores?

brotskydotcom: Thanks for the review. I agree this is messy.  Unfortunately, in my experience, dealing with edge cases is always messy, and (as I explain further in #214) having more than one keystore enabled on a given platform is a fairly rare edge case. Using features to specify a single (default) keystore per platform allows the vast majority of clients to write platform-independent code.

As to the question about what this will look like if we integrate a whole bunch of new keystores on one platform, I'm not particularly concerned that's going to happen.  The only reason the `keyutils` platform was integrated in the first place was to allow using this crate on headless Linux boxes where the SecretStore is generally not available.

Hope this helps explain my reasoning.

soywod: > The only reason the `keyutils` platform was integrated in the first place was to allow using this crate on headless Linux boxes where the SecretStore is generally not available.

Could you explain a bit more, or give an example? I am not sure to understand.

soywod:
I propose a lighter version in https://github.com/hwchen/keyring-rs/pull/220.

src/lib.rs
brotskydotcom: The secret service relies on the system to prompt users out of band with their current process, whenever it encounters a locked item or collection. Headless linux installs don't support this, so secret-service can only be used by unlocking the default collection when starting the service (see the docs in the secret-service module for more about this).

Because @landhb and I wanted to have a default keystore available on headless Linux boxes, he contributed and I integrated the keyutils implementation. There was never any expectation that the both would be used at the same time, because if the secret service can be used there's no reason to use the keyutils.

soywod: Understood, thanks for the details. From the keyutils inline doc:

> Consider the keyring a secure cache and plan for your application to handle cases where the entry is no-longer available in-memory.

For this particular (and unique) case, having both the keyutils and secret service makes sense (hence #222), right?

brotskydotcom: Absolutely! See my review comments on #222, which I think would be a terrific addition to this crate (done as a separate "combo" keystore rather than as a modification of the existing keystore).

brotskydotcom:
Based on discussion in various issues, this PR now makes secret-service the default if both it and keyutils are enabled.

soywod:
Looks great, I guess it can be merged!

soywod (approved):

PR #220, opened 2024-10-11
Improve secret service and keyutils compile error condition

soywod:
Fixes #214 #215.

This PR adjusts the compilation error condition so that `{async,sync}-secret-service` and `keyutils` can be used together but not `async-secret-service` and `sync-secret-service`.

Issue #168, opened 2024-04-28
Is it possible to tell if secure storage is available before storing something?


John-Nagle:
I'd like to know whether secure storage is available before trying to store anything. On some Linux systems, no secure storage seems to be available.

Use case: deciding whether to display a "Remember password" checkbox. 

Thanks.

brotskydotcom:
Hi @John-Nagle, thanks for the question.

If I understand you correctly, you are asking whether there's a way to find out whether the lifetime of a stored credential survives a reboot.  This is similar to #147, and the sample code in that issue shows you how to use the `persistence` method on a credential store to answer that question.  Please let me know if this gives you what you need.

John-Nagle:
Ah, good.

PR #222, opened 2024-10-11
Provide keyutils with persistence-after-reboot using secret-service

soywod:
This PR makes both async and sync secret service keystores using keyutils as cache, when available (on Linux only).

Since you mentioned that rare are users needing both secret service *and* keyutils, I assume the following statement: if a user enables both secret service and keyring, it implies that the user wants to benefit from the Linux secure cache. To disable cache, just remove the feature.


soywod:
Requires #220 for the CI not to complain about dual keystores.

.github/workflows/ci.yaml
```diff
@@ -46,6 +46,7 @@ jobs:
       matrix:
         features:
           - "sync-secret-service"
+          - "sync-secret-service,linux-native"
```
brotskydotcom, line 49: This additional test is a good one, and needs to be part of the CI once #215 (or #220) is integrated.  Thanks for pointing that out!

brotskydotcom (changes requested):
Hi Clément, thanks for this! It's a very interesting idea.  I have reactions at three levels:

At the highest level, I think this captures very nicely the idea you were talking about in one of your comments, where you talked about having one credential store be primary and then another be fallback. That idea is really independent of credential store, and could be a separate crate that uses keyring to provide the credential stores.

At an intermediate level, this feels like it should be structured as a separate credential store module (and feature) that relies on (and requires) both secret-service and keyutils. Rather than thinking of it as "secret-service with keyutils cache," I tend to think of it as "keyutils with fallback to secret-service for persistence across reboots," because that's how it would work in practice (as has been noted in a number of closed issues against this crate, for example see #168).

At the lowest level, I think there are some logic changes that would be needed to really make this robust, in particular one would need to guard against the two stores going "out of sync" with one another as a result of separate access.

If you are willing to turn this into a separate credential module, and fix the out-of-sync issues, I would be delighted to include it as part of keyring.

soywod:
> At an intermediate level, this feels like it should be structured as a separate credential store module (and feature) that relies on (and requires) both secret-service and keyutils. Rather than thinking of it as "secret-service with keyutils cache," I tend to think of it as "keyutils with fallback to secret-service for persistence across reboots," because that's how it would work in practice (as has been noted in a number of closed issues against this crate, for example see #168).

I thought about it, I hesitated between the actual simpler (naive) solution and this one. What stopped me to implement the late one is the complexity around: how to name such store? how to deal with cargo features (sync/async secret service with/without keyring)? it brings even more conditions etc. I am not against, but knowing the low need of such feature we could start simpler, by just adding cache to secret service. It can be put behind a feature gate as well.

> At the lowest level, I think there are some logic changes that would be needed to really make this robust, in particular one would need to guard against the two stores going "out of sync" with one another as a result of separate access.

Same here. While I am not against (because it makes sense), I find overkill to fine-tune the feature now knowing that it's an edge case of an edge case.

IMHO, better to start simpler and improve over time with users' feedback. What do you think?

src/secret_service.rs
```diff
@@ -174,6 +192,13 @@ impl CredentialApi for SsCredential {
     /// returns an [Ambiguous](ErrorCode::Ambiguous)
     /// error with a credential for each matching item.
     fn get_password(&self) -> Result<String> {
+        #[cfg(all(target_os = "linux", feature = "linux-native"))]
+        if let Some(keyutils) = &self.keyutils {
+            if let Ok(password) = keyutils.get_password() {
+                return Ok(password);
```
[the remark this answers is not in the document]
soywod: As I mentioned in my previous comment, it's already an edge case (both secret service and keyutils) of an edge case (passwords not sync). I would rather remain simple and explain in the doc that there is no integrity check at the moment. Chances that someone decide to manually change the password in the keyutils and not in the secret service are low. Worst case, the user just have to reboot his computer to clear the keyutils cache.

I already use this basic mechanism in my app and it works well enough so far. The aim is really just to retrieve faster and safely a password that had already been asked recently, during the current same session.

soywod:
After all, both ideas can be complementary, as they solve different issues:

- A dedicated store for secret service + keyutils makes sense if you want to have a solid, safe and secure cache system. Integrity matters, in this case I agree with all your comments.
- A keyutils as light cache system for secret service makes sense in opportunistic way: if you already asked for a password that is in the cache → great, use it, otherwise it's fine → use secret service

Since nobody needed both secret service and keyutils (nor such cache feature) before, it makes sense to start *first* with the light one.

brotskydotcom:
Thanks for your thoughtful comments; they make a lot of sense.

I absolutely agree that keeping things simple to begin with is the way to go. It doesn't need to be super robust as long as we document what the behavior is.

I agree there's a real need to have a standard way for keyring to support backing the keyutils with persistence. And I think that fills a real need for people who have systems they run as headless servers for the most part (relying on keyutils) but then can log into via UI to configure (via secret-service).

I still think this is best modeled as a separate linux-only keystore (perhaps called "linux-native-persistent") that is an exclusive alternative to both keyutils and secret-service. Because both the keyutils and dbus-secret-service modules are optional dependencies, we can have the "linux-native-persistent" feature enable both of those modules, and then simply use them as a client (using the pattern you've established in your existing PR). That way we don't complicate the existing keystore code and we give all users an easy way to use the double-keystore pattern that you are using.

Is this an approach that might work for you?

soywod:
Maybe `linux-native` makes even more sense than `linux-native-persistent`: it aligns with other `apple-native` and `windows-native` (providing a persistent store). `keyutils` and `{sync,async}-secret-service` could still be used independently if needed. Let me update the PR tomorrow morning so we can discuss on it. Thank you!

soywod:
I had hard time to deal with this new keystore with the actual cargo features management. As I expected, it became a mess of conditions in order to make default keystore exclusive.

I also discovered the potential reason behind the default keystore confusion: the word `native` *implies* that the current store is available, by default, on the target machine. `linux-native(-persistent)` is not really native because it depends now on other things like secret service or dbus.

In order to go forward, I suggest the following ideas:

- `*-native` cargo features stand for their meaning: available natively, without any other deps, on the target OS:
  - `linux-native` implies keyutils
  - `apple-native` implies security-framework
  - `windows-native` implies windows-sys
- `default-*` cargo features stand for *persistent* keystore that should be use by default on the target OS:
  - `default-linux` implies the new secret service + keyutils store
  - `default-linux-headless` implies the keyutils only (the only one not persistent)
  - `default-bsd` implies the secret service only
  - `default-apple` implies security-framework
  - `default-windows` implies windows-sys

`default-linux` and `default-bsd` need an explicit `{sync,async}-secret-service` feature to work, this way it prevents duplicate cargo features like `default-linux-{sync,async}-secret-service` and it makes [compilation conditions](https://github.com/hwchen/keyring-rs/pull/222/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759L218-R234) easier to understand.

brotskydotcom (changes requested):
This is starting to look very good, thanks for working on it!  Some notes:

1. WRT your suggested feature renames, changing existing feature behavior is a SemVer-incompatible change, and I would like to keep this change to a backwards-compatible enhancement (dot) release.  So for now let's pick a new name for the feature controlling this keystore and just add it alongside the existing ones as an exclusive option. I know this makes the `cfg` logic even deeper on the `*nix` side, but that's purely an internal issue that we can manage and doesn't show to clients.
2. I like where you're going with the idea about `-default` vs `-native` features, and will keep track of it for the next major release. It also suggests to me that we don't need the `-native` feature set at all, because we can just include those keystores based on platform rather than feature, so maybe this will all get a lot simpler in v4 :).  After all, there is only 1 platform (*nix) where there's a choice of keystores...
3. I don't see any reason to support both sync and async secret-service usage in this brand new keystore. Just supporting the `dbus-secret-service` will provide a lot of advantages: (a) no need to conditionalize a bunch of code, (b) much simpler feature logic (just have this feature rely on `keyutils-rs` and `dbus-secret-service` directly and prevent use of other *nix keystores), (c) much slimmer dependency footprint.
4. I think we should consider altering (or allowing the client to alter) one or more of the target, user, and service values that this code passes to keyutils and secret-service in order to avoid conflicts with legacy code that was already using those services for their passwords. But then again maybe we want this code just to overtake all those old entries if they exist?
5. When get_secret gets a miss in keyutils and then a hit in secret-service, it should repair the cache with set_password on the keyutils entry. If we do decide to overtake old entries with the same values, then perhaps the same should be done if we get a hit in keyutils but a miss in secret-service.

soywod:
> 3. I don't see any reason to support both sync and async secret-service usage in this brand new keystore. Just supporting the dbus-secret-service will provide a lot of advantages: (a) no need to conditionalize a bunch of code, (b) much simpler feature logic (just have this feature rely on keyutils-rs and dbus-secret-service directly and prevent use of other *nix keystores), (c) much slimmer dependency footprint.

And what if someone already relies on zbus (it may be my case within my app, I need to check) and want to have `keyutils` + zbus secret service in order to avoid having both dbus and zbus as dependencies?
   
> 4. I think we should consider altering (or allowing the client to alter) one or more of the target, user, and service values that this code passes to keyutils and secret-service in order to avoid conflicts with legacy code that was already using those services for their passwords. But then again maybe we want this code just to overtake all those old entries if they exist?

I think it's fine to overtake old entries, at least for now. Let's wait for users feedback?

> 5. When get_secret gets a miss in keyutils and then a hit in secret-service, it should repair the cache with set_password on the keyutils entry. If we do decide to overtake old entries with the same values, then perhaps the same should be done if we get a hit in keyutils but a miss in secret-service.

Nice catch, I take care of it tomorrow morning.

---

I don't get the reason of the [CI failure](https://github.com/hwchen/keyring-rs/actions/runs/11313329332/job/31467169100?pr=222), nothing changed except for cargo features. Any idea?

brotskydotcom:
Wrt using dbus-secret-service with tokio and zbus: I have tested that thoroughly and there's no issue there. I'd really like to keep the footprint as light as possible and I don't want to have multiple features for this new store, so let's stay with just the one.

Not sure about the test failures, often they fail once randomly but when I ran them again they failed at the same test.  Once you finish your changes we can take a closer look.

soywod:
I just pushed changes. Somehow, by keeping the `secret-service` feature guard it makes conditions lighter. Plus it enables the ability to use both keyutils with zbus secret service (which is still a plus for me at application level, I don't want to rely on external lib like libdbus). And it looks like it does not break the current API, do you confirm?

> 2. I like where you're going with the idea about -default vs -native features, and will keep track of it for the next major release. It also suggests to me that we don't need the -native feature set at all, because we can just include those keystores based on platform rather than feature, so maybe this will all get a lot simpler in v4 :). After all, there is only 1 platform (*nix) where there's a choice of keystores...

I thought about this, but native features are still convenient for testing purpose (for example to force the mock store). It joins a bit the idea I proposed last time about having the whole default keystore system behind a feature gate (on by default).

Native features also allow apps to choose on which OS they want to be compatible: if you have an app that works only on Linux and MacOS, you can omit the `windows-native` which will make the crate not buildable on Windows (better that still building but app failing at runtime).

> 5. If we do decide to overtake old entries with the same values, then perhaps the same should be done if we get a hit in keyutils but a miss in secret-service.

If we get a hit in keyutils then it returns straight without hitting secret service: the role if this cache is to prevent as much as possible to call the slower secret service. It's part of the opportunistic approach I was explaining, it's fine (at least with this particular store) if keys are not sync.

src/secret_service_with_keyutils.rs
```diff
@@ -0,0 +1,169 @@
...
+        let _ = self.keyutils.set_secret(secret);
+        Ok(())
+    }
+
+    fn get_password(&self) -> Result<String> {
+        if let Ok(password) = self.keyutils.get_password() {
+            return Ok(password);
+        }
+
+        let password = self.ss.get_password()?;
```
brotskydotcom, line 44: This passes errors from the secret service directly to the client, which you can't do, because the secret service the Ambiguous error, which this keystore must not (see comments below), and also because that Ambiguous error would contain secret-service credentials, not credentials from this service (see comments below). You have to analyze the error returned and return an appropriate error instead, which should only be errors that don't return credentials.

src/secret_service_with_keyutils.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+impl CredentialApi for SsKeyutilsCredential {
+    fn set_password(&self, password: &str) -> Result<()> {
+        self.set_secret(password.as_bytes())
+    }
+
+    fn set_secret(&self, secret: &[u8]) -> Result<()> {
+        self.ss.set_secret(secret)?;
+        let _ = self.keyutils.set_secret(secret);
```
brotskydotcom, line 35: You cannot ignore this error, because there are secrets that are acceptable to secret service (such as an empty secret) that are not acceptable to keyutils. If your service is going to cache in keyutils, it should only accept secrets that can be cached in keyutils. I would recommend that you invert the logic here: first store in keyutils, and then if you don't get an error store in secret service.

src/secret_service_with_keyutils.rs
```diff
@@ -0,0 +1,169 @@
...
+        Ok(Self { keyutils, ss })
+    }
+
+    pub fn new_from_item(item: &Item) -> Result<Self> {
+        let ss = SsCredential::new_from_item(item)?;
+        let keyutils = KeyutilsCredential::new_from_item(item)?;
+        Ok(Self { keyutils, ss })
+    }
+
+    pub fn get_all_passwords(&self) -> Result<Vec<String>> {
```
brotskydotcom, line 127: remove this function, it's not needed. It exists in secret-service only to help deal with ambiguity.

src/secret_service_with_keyutils.rs
```diff
@@ -0,0 +1,169 @@
...
+        self.ss.get_all_passwords()
+    }
+
+    pub fn delete_all_passwords(&self) -> Result<()> {
```
brotskydotcom, line 131: remove this function, it's not needed. It exists in secret-service only to help deal with ambiguity.

src/secret_service_with_keyutils.rs
```diff
@@ -0,0 +1,169 @@
...
+        self.ss.delete_all_passwords()
+    }
+
+    pub fn all_attributes(&self) -> HashMap<&str, &str> {
```
brotskydotcom, line 135: Remove this function, it's never called, because the secret service is a concrete implementation and you can't override it.

src/secret_service_with_keyutils.rs
```diff
@@ -0,0 +1,169 @@
...
+        self.ss.all_attributes()
+    }
+
+    pub fn search_attributes(&self, omit_target: bool) -> HashMap<&str, &str> {
```
brotskydotcom, line 139: Remove this function, it's never called, because the secret service is a concrete implementation and you can't override it.

Cargo.toml
```diff
@@ -17,8 +17,11 @@ linux-native = ["dep:linux-keyutils"]
 apple-native = ["dep:security-framework"]
 windows-native = ["dep:windows-sys", "dep:byteorder"]
 
-sync-secret-service = ["dep:dbus-secret-service"]
-async-secret-service = ["dep:secret-service", "dep:zbus"]
+keyutils = ["dep:linux-keyutils"]
+secret-service = []
```
brotskydotcom, line 21: I don't like the introduction of these "do-nothing" features.  It seems you have introduced them just to make the configuration logic look simpler, but the configuration logic is not visible to clients and these features are.

Cargo.toml
```diff
@@ -17,8 +17,11 @@ linux-native = ["dep:linux-keyutils"]
...
+secret-service-with-keyutils = ["secret-service", "keyutils"]
```
brotskydotcom, line 22: If you insist on support both sync and async, then just have two features "sync-secret-service-with-keyutils" and "async-secret-service-with-keyutils", and have them enable the appropriate dependencies. So you have to add a couple of "any" conditions in the cfg logic.  Big deal. If you're adding two keystores (which you are), then have a feature for each one, and be done with it. It's the client view of the features that should be simple, not the configuration view.

brotskydotcom (changes requested):
OK, this is getting there.  Thanks for all your work.  Here are some top-level comments, and then there are lots of detailed comments in the files.

I think we need more clarity about what exactly this keystore is. From my perspective, there are two basic ways to think about this keystore:

1. It's the secret service with "fast, in-memory caching" provided by keyutils, where the cache only lasts until reboot. (I believe this is how you think of it.) I would argue that there's not much utility in this, because credential stores are not expected to be fast.
2. It's the keyutils store with entry persistence beyond reboot provided by the secret service. (This is how I think of it.) This has tremendous utility because it means you can run headless and unlock the secret store once at startup (which can be done in PAM or by asking the user for the secret store password) to restore the persistent values.  (Alternatively, you can run a headed process to restore or update the persistent values.)

The difference between (1) and (2) is not so much in how it works as in what it accepts and what it exposes. If we think of this service primarily as keyutils plus persistence, then we can only accept passwords that can be stored in keyutils, entries have no additional attributes, and there is no ambiguity. If we think of this service primarily as secret-service plus a cache, then there are entries that cannot be cached, entries have additional attributes, and there is ambiguity. But if clients really are going to want those characteristics, then they should just use secret service and ignore keyutils, because none of those other characteristics can be cached (at least not in any simple implementation).

Approaching this from a different perspective: whenever you compose two services, what you get is a service that only accepts the "greatest common denominator" of the inputs and produces the "greatest common denominator" of the outputs. In this case, keyutils itself is the greatest common denominator: all of its inputs are acceptable to secret service and it has no outputs other than passwords. Trying to do more than that just leads to problems, because the cache can't handle what the persistent store can.

I think we need to think of this keystore as being "keyutils plus persistence" (which is where my original name came from). That will drastically simplify the implementation. Once we get to that point then we can add docs and tests.

Let me know what you think.

src/keyutils_persistent.rs
```diff
@@ -0,0 +1,140 @@
...
+        let Ok(()) = self.ss.update_attributes(attributes) else {
+            return self.keyutils.update_attributes(attributes);
+        };
+
+        Ok(())
+    }
+
+    fn delete_credential(&self) -> Result<()> {
+        self.keyutils.delete_credential()?;
+        self.ss.delete_credential()?;
```
brotskydotcom, line 80: This doesn't work if the keyutils entry has been cleared (e.g., by a reboot), but the secret service credential exists. An explicit delete should try to delete both credentials and only return "doesn't exist" if that error was returned by both.

src/keyutils_persistent.rs
```diff
@@ -0,0 +1,140 @@
...
+    }
+}
+
+impl KeyutilsPersistentCredential {
+    pub fn new_with_target(target: Option<&str>, service: &str, user: &str) -> Result<Self> {
+        let ss = SsCredential::new_with_target(target, service, user)?;
+        let keyutils = KeyutilsCredential::new_with_target(target, service, user)?;
+        Ok(Self { keyutils, ss })
+    }
+}
```
brotskydotcom, line 99: Excellent!

src/lib.rs
```diff
@@ -182,50 +182,69 @@ pub mod mock;
...
 //
-#[cfg(all(feature = "sync-secret-service", feature = "async-secret-service"))]
+#[cfg(any(
+    all(feature = "sync-secret-service", feature = "async-secret-service"),
+    all(
+        feature = "sync-persistent-keyutils",
+        feature = "async-persistent-keyutils",
+    )
+))]
 compile_error!("This crate cannot use the secret-service both synchronously and asynchronously");
```
brotskydotcom, line 192: How about "This crate cannot use both the sync and async versions of any credential store"

src/lib.rs
```diff
@@ -182,50 +182,69 @@ pub mod mock;
...
 
 //
 // pick the *nix keystore
 //
-
-#[cfg(all(target_os = "linux", feature = "linux-native"))]
+#[cfg(all(target_os = "linux", feature = "keyutils"))]
```
brotskydotcom, line 197: This change will be reverted when you remove the keyutils feature

src/lib.rs
```diff
@@ -182,50 +182,69 @@ pub mod mock;
...
 pub mod keyutils;
-// use keyutils as default if secret-service is not available
 #[cfg(all(
     target_os = "linux",
     feature = "linux-native",
-    not(any(feature = "sync-secret-service", feature = "async-secret-service"))
+    not(feature = "sync-secret-service"),
+    not(feature = "async-secret-service"),
 ))]
```
brotskydotcom, line 204: I like the unrolling of the not-any!! And now I see why you wanted to separate linux-native from keyutils.  But that's not necessary, because if someone has enabled either version of the persistent-keyutils store they will also have enabled one of the secret-service features, so keyutils will still not be chosen as the default.

src/lib.rs
```diff
@@ -182,50 +182,69 @@ pub mod mock;
...
+
+// fallback to mock if neither keyutils nor secret service is available
+#[cfg(all(
+    any(target_os = "linux", target_os = "freebsd", target_os = "openbsd"),
+    not(any(
+        feature = "keyutils",
+        feature = "sync-secret-service",
+        feature = "async-secret-service",
+    )),
+))]
```
brotskydotcom, line 247: Revert this back to linux-native

Cargo.toml
```diff
@@ -13,10 +13,13 @@ exclude = [".github/"]
 readme = "README.md"
 
 [features]
-linux-native = ["dep:linux-keyutils"]
+linux-native = ["keyutils"]
 apple-native = ["dep:security-framework"]
 windows-native = ["dep:windows-sys", "dep:byteorder"]
 
+keyutils = ["dep:linux-keyutils"]
```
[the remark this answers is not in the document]
soywod: It was just to align with other linux features. Since you have `{sync,async}-secret-service` (and not `linux-secret-service`), I thought `keyutils` would align nicely. Especially that we have now `{sync,async}-persistent-keyutils`.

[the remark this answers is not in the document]
brotskydotcom: Ah, yes, I see your point.  But I'd rather fix that by calling the new keystore `linux-native-persistent` rather than `linux-keyutils-persistent`. The reason there's no linux- in front of secret-service is that it's not linux-only.  keyutils *is* linux-only, and built in, which is why I called it linux-native.

src/keyutils_persistent.rs
```diff
@@ -0,0 +1,140 @@
...
+
+        let secret = self.ss.get_secret().map_err(ambigous_to_no_entry)?;
+        self.keyutils.set_secret(&secret)?;
+
+        Ok(secret)
+    }
+
+    fn get_attributes(&self) -> Result<HashMap<String, String>> {
+        let Ok(attributes) = self.ss.get_attributes() else {
+            return self.keyutils.get_attributes();
```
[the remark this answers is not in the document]
soywod: Now that we took the direction of the persistent keyutils rather than secret service with cache, I would say that it's not necessary to implement attributes-related function (like for the keyutils credential). Let me remove them.

src/lib.rs
soywod: Wait, it will not work as expected if someone add the `linux-native` and one of the secret service?

soywod: I confirm it does not work, see <https://github.com/soywod/keyring-rs/actions/runs/11353166029/job/31577538792>. I revert the changes.

soywod:
The condition you were proposing did not work, so I adjusted them. Ready for the next review!

Cargo.toml
```diff
@@ -17,6 +17,8 @@ linux-native = ["dep:linux-keyutils"]
 apple-native = ["dep:security-framework"]
 windows-native = ["dep:windows-sys", "dep:byteorder"]
 
+sync-persistent-keyutils = ["dep:linux-keyutils", "sync-secret-service"]
```
brotskydotcom, line 20: should be:
```
sync-persistent-keyutils = ["linux-native", "sync-secret-service"]
```
and same change for async.  This is what I meant and what makes the original config work for loading keyutils and choosing the default.

It makes even more sense if you rename this feature to "linux-native-sync-persistent" and the next one to "linux-native-async-persistent".

src/keyutils_persistent.rs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        let secret = self.ss.get_secret().map_err(ambigous_to_no_entry)?;
+        self.keyutils.set_secret(&secret)?;
+
+        Ok(secret)
+    }
+
+    fn delete_credential(&self) -> Result<()> {
+        // TODO: log the error
```
brotskydotcom, line 61: OK this works well enough as long as we log the error - good argument for logging! :)  Please add the log entries wherever we ignore a result once you rebase on top of the merged master.

src/lib.rs
```diff
@@ -182,50 +182,74 @@ pub mod mock;
...
+    )
+))]
+compile_error!("This crate cannot use both the sync and async versions of any credential store");
 
 //
 // pick the *nix keystore
 //
-
-#[cfg(all(target_os = "linux", feature = "linux-native"))]
+#[cfg(all(
```
brotskydotcom, line 197: Revert this once you make the feature change noted in Cargo.toml (having the new stores depend on linux-native.

brotskydotcom (changes requested):
Excellent! One last feature name change (or three if you like my suggested change to the top-level features names) and adding logging and the content is done!

Now we need documentation and unit tests.  You should pretty much be able to crib the tests from the keyring crate. If you can take a first pass at the docs - I suggest keeping them simple and just referring to the docs of the other crates if necessary - I am happy to do copy editing in a review.

soywod:
> I am happy to do copy editing in a review

Sure, maybe it is better this way as I may not word properly things. You can also push a commit directly on my branch, as you wish.

Ready for another review.

brotskydotcom (changes requested):
Almost! You lost a config test that's required for non-linux systems.  I've called it out. (actually, I didn't before, I will add it below.)

What do you think about adding logging for all the error conditions that aren't returned.  For example, when you miss in keyutils we would log that happened even if you then hit in secret-service. As long as we have logging I think it would be good to be transparent in this way.

src/lib.rs
```diff
@@ -184,50 +184,67 @@ pub mod mock;
...
+))]
+pub use keyutils_persistent as default;
+
+// fallback to mock if neither keyutils nor secret service is available
+#[cfg(all(
+    any(target_os = "linux", target_os = "freebsd", target_os = "openbsd"),
+    not(feature = "linux-native"),
+    not(feature = "sync-secret-service"),
+    not(feature = "async-secret-service"),
+))]
```
brotskydotcom, line 247: You lost this or'd test in the choice of mock as default:
```
    all(
         any(target_os = "freebsd", target_os = "openbsd"),
         not(any(feature = "sync-secret-service", feature = "async-secret-service"))
```
That's an important one because someone may have specified linux-native as their keystore but then be building on a non-linux platform - in that case they use mock as default.

soywod:
Not sure to understand what's wrong with the CI. Feel free to push commits on the branch if needed for the merge, I will check tomorrow morning.

brotskydotcom:
Don't worry about the test failures until we have unit tests and docs.  Then, if necessary, I'll pull it and look at it.

soywod:
I initiated doc and unit tests, feel free to adjust at your convenience.

brotskydotcom (approved):
Great, thank you!!  I will be making tweaks for a while by pushing commits to your repo - feel free to review them.  Then once we agree it's done I'll merge it.

brotskydotcom:
OK, I've updated docs and tweaked a few things.  Please take a look and try it out and let me know if you think we are ready to pull this.

soywod:
Could I ask you to release a new version as soon as you merge? It will allow me to release my versions as well.

PR #223, opened 2024-10-11
Add logs using `logs`

### LINKED ISSUE #221: Add logs support
It could be a great and useful addition to add the [`tracing`](https://docs.rs/tracing/latest/tracing/) crate behind a feature gate and to emit logs. I can propose a PR if you are interested in seach feature.

soywod:
Closes #221.

This PR adds basic log support using the [`tracing`](https://github.com/tokio-rs/tracing) crate. It gives more context to users about what is going on inside `keyring-rs`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,7 +37,7 @@ jobs:
       - name: Build and Test
         env:
           RUST_LOG: debug
-        run: cargo test --features=apple-native,windows-native --verbose -- --nocapture
+        run: cargo test --features=apple-native,windows-native --verbose
 
       - name: Build the CLI release
         run: cargo build --release --features=apple-native,windows-native --example keyring-cli
@@ -90,7 +90,7 @@ jobs:
         env:
           RUST_LOG: debug
         # run tests single-threaded to avoid dbus race conditions
-        run: cargo test --features=${{ matrix.features }} -- --test-threads=1 --nocapture
+        run: cargo test --features=${{ matrix.features }} -- --test-threads=1
 
   ios_native:
     runs-on: macos-latest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,7 +35,9 @@ jobs:
         run: cargo clippy -- -D warnings
 
       - name: Build and Test
-        run: cargo test --features=apple-native,windows-native --verbose
+        env:
+          RUST_LOG: debug
+        run: cargo test --features=apple-native,windows-native --verbose -- --nocapture
 
       - name: Build the CLI release
         run: cargo build --release --features=apple-native,windows-native --example keyring-cli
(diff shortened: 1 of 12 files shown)


soywod:
More logs can be added if #222 is accepted and merged, so I leave this PR in draft for now.

brotskydotcom (changes requested):
1. If you change the dependency to be on the `logs` crate, then you won't have to conditionalize the logging code at all, which will provide an enormous simplification. (See my comments on #221 for other reasons I prefer logs to tracing.)
2. When creating an entry, please be sure to log the credential-store-specific attributes being used as well as the credential store. (Since all of the stores derive Debug, you can just debug print the credential into the log.)
3. When operating on an entry, be sure to debug log the entry. That way it can be correlated to where the entry was created.
4. I suggest you move logging of the target, service, and user into the platform-independent code, which saves you adding it to the credential-store code.  Then the credential-store code can just debug log the credential being operated on.

soywod:
The PR is ready for review. In fact it can be mergde before or after #222, it does not matter.

I try to apply what you asked for, let me know if it's what you have in mind. I added `env_logger` for integration tests, so you can easily get logs with `RUST_LOG=debug`. I enabled it for the CI so we have logs always available, in case of. Just a proposition.

.github/workflows/ci.yaml
```diff
@@ -35,7 +35,9 @@ jobs:
         run: cargo clippy -- -D warnings
 
       - name: Build and Test
-        run: cargo test --features=apple-native,windows-native --verbose
+        env:
+          RUST_LOG: debug
+        run: cargo test --features=apple-native,windows-native --verbose -- --nocapture
```
brotskydotcom, line 40: why `--nocapture`? You don't seem to be verifying the logs. In a CI context, this seems to make the logs unnecessarily long.

IIRC, even without --nocapture, if a test fails then the debug log output will be shown for that test.

.github/workflows/ci.yaml
```diff
@@ -85,8 +87,10 @@ jobs:
         run: gnome-keyring-daemon --components=secrets --daemonize --unlock <<< 'foobar'
 
       - name: Run tests
+        env:
+          RUST_LOG: debug
         # run tests single-threaded to avoid dbus race conditions
-        run: cargo test --features=${{ matrix.features }} -- --test-threads=1
+        run: cargo test --features=${{ matrix.features }} -- --test-threads=1 --nocapture
```
brotskydotcom, line 93: same question as above

brotskydotcom (changes requested):
This looks very good, thank you! Minor language cleanups and format cleanups and a question for you to consider.

src/lib.rs
```diff
@@ -328,18 +329,25 @@ impl Entry {
...
+        debug!("created entry with {:?}", entry.inner);
+        Ok(entry)
     }
 
     /// Create an entry for the given target, service, and user.
     ///
     /// The default credential builder is used.
     pub fn new_with_target(target: &str, service: &str, user: &str) -> Result<Entry> {
-        build_default_credential(Some(target), service, user)
+        debug!("creating entry with service {service}, user {user} and target {target:?}");
```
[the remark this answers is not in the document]
soywod: My bad, I thought it was an `Option<&str>`. Fixed!

.github/workflows/ci.yaml
soywod: It was to force them to show, just a suggestion. I removed it.

brotskydotcom (approved):
Perfect! Thanks so much for this! I will merge it right away, so you can rebase #222 on it.

Issue #235, opened 2025-02-24
Doc: missing for some modules


unkcpz:
In https://docs.rs/keyring/latest/keyring/#modules, it only has the secret-service module where I can go and check the doc of credential description. 
I was expect also the doc for the keyutils and for ios/macos. 

I assume [build-xplat-docs.sh](https://github.com/hwchen/keyring-rs/blob/master/build-xplat-docs.sh) is for building cross-platform doc. But the build OS is linux so the apple part is ignored. 
I believe [this is a more standard way to build doc with cross-platform doc](https://doc.rust-lang.org/rustdoc/advanced-features.html#cfgdoc-documenting-platform-specific-or-feature-specific-information)?

brotskydotcom:
Hi Jusong, thanks for raising this question. Because I've assumed that people only need docs for the platforms they can actually build for, and because docs.rs builds for all the "common" platforms (including macOS, although not iOS), I've never thought about doing an "all-platforms" build of the doc. I would be happy to look at a PR if you're interested in providing one.

unkcpz:
It is mostly for the python binding [keyringrs](https://github.com/unkcpz/keyringrs) you may notice through [another issue](https://github.com/hwchen/keyring-rs/issues/234) I opened. I want to redirect from the readme to docs.rs for all platforms for options such as definition of `target` option of different platform.

I am surely willing to fire a PR.  Could you tell me is `build-xplot-docs.sh` is used or it can replace with trying to use `#[cfg(doc)]`.

brotskydotcom:
The `build-xplat-docs.sh` file is just for local development purposes and should be removed as part of your PR. By all means try getting the docs working using `#[cfg(doc)]`

brotskydotcom:
The `build-xplat-docs.sh` file is just for local development purposes and should be removed as part of your PR. By all means try getting the docs working using `#[cfg(doc)]`

unkcpz:
thanks, I draft a PR in #237, let's move the discussion there. It actually not trivial to make docs build for cross-platfrom, which regard as [the ['holy grail of rustdoc'](https://github.com/rust-lang/rust/issues/1998)](https://doc.rust-lang.org/rustdoc/advanced-features.html?search=dox) :)

Issue #227, opened 2024-10-19
List of real-world apps using `keyring-rs`


soywod:
If there any place listing apps based on `keyring-rs`? If not then I would love to propose such section in the README. It can be useful for new users, and it can make your lib more visible as well. I already have few libs and apps to list there.

brotskydotcom:
Great idea. Let's turn this issue into a place where anyone who has an app or library they want listed can add a comment giving the name, brief description, and (optionally) where to find the app (an URL). Then every time we cut a new release we will update the README with the list of apps collected so far.

So, clients, please add a comment here giving the info about your client(s)! Thanks!

soywod:
Awesome, then I will start posting my libs and apps using `keyring-rs` as soon as I release them. It could be great to also pin this issue so that it is more visible for new commers.

soywod:
[`keyring-lib`](https://crates.io/crates/keyring-lib): a High-level, asynchronous API for `keyring-rs`.

The aim of this crate is just to wrap `keyring-rs` with sane (opinionated) defaults: async with `tokio` or `async-std`, error management using [`thiserror`](https://docs.rs/thiserror/latest/thiserror/), global service name to prevent redundancy etc. It also serves as a gateway to other libs and apps of my project, which I will share soon as well.

soywod:
[**secret-lib**](https://crates.io/crates/secret-lib): *Cross-platform, asynchronous Rust library to retrieve secrets from different sources*

The aim of this crate is to be able to retrieve secrets using different backends (including one based on `keyring-rs`). It is also opinionated, and proposes an async API with error management based on [`thiserror`](https://docs.rs/thiserror/latest/thiserror/).

soywod:
[**mml-lib**](https://crates.io/crates/mml-lib): *Rust implementation of the Emacs MIME message Meta Language, as known as [MML](https://www.gnu.org/software/emacs/manual/html_node/emacs-mime/Composing.html)*

`keyring-rs` is used inside this project to save secret keys and passphrases when used with the PGP native feature.

pk8189:
@brotskydotcom I'm hoping to use this lib for my API client CLI that stores user API keys for future use. Would you recommend this library for that use case? Will give a shout out in our repo if you think it makes sense!

brotskydotcom:
> @brotskydotcom I'm hoping to use this lib for my API client CLI that stores user API keys for future use. Would you recommend this library for that use case?

[...]

PR #237, opened 2025-02-25
doc: cross-platform with target tags

unkcpz:
ref #235

The goal of thi PR are:
- [x] Showing `keyutils` and `keyutils_persistent` module doc in platform dropdown of x86_64_linux target (see fig below). 
- [x] tag for target OS so it is clear in what condition is the module avail.
- [x] module section for apple-ios
- [x] CI doc test for cross-platform
- [ ] ~~portal list redirect to platform links~~ (may not worth to do it)

image

The plans are to make the docs more friendly to easy redirect to correspond platform is build for each platform with all module section existed (the keyutils is missing for the linux at the moment). Then in the index page of the doc, having a list as the portal to different platform. 

Allowed docs.rs to build for multiple OS.
The experimental doc_cfg feature is used to tag module for its targeting OS.

The caveat of this approach is not sure the docs.rs can properly display docs for all platform (I'll test with using docs.rs docker-compose to try locally, so I put it as draft). From what documented in https://docs.rs/about/metadata, and reference how it did in [CI of security-framework crate](https://github.com/kornelski/rust-security-framework/blob/d0b1b60afc0a000f5f008a70d0e7e55362bae79a/security-framework/Cargo.toml#L67-L69). I think it will work. 

I keep the `build-xplot-docs.sh` so it can be used to test the docs can be build for corresponding OS. It can maybe also server as the script for docs build test, if the approach of this PR adapted. 

UPDATE: apparently after test using docs.rs docker-compose locally, using `targets` in `[package.metadata.docs.rs]` is **NOT** to have a single platform agnostic page to show all, but to have a platform dropdown that user can select which doc to look at (which is already there for the crate but I didn't realize, sorry about it). (see https://github.com/rust-lang/docs.rs/issues/2749 for discussion)

Note for local docs.rs test:
- https://forge.rust-lang.org/docs-rs/add-dependencies.html is required for the pkg-config dependency.
- https://github.com/rust-lang/docs.rs


unkcpz:
@brotskydotcom The pr is ready to be review. 

For the last checkbox, maybe there is a good way to have some thing in the index crate docstring which I don't know (can follow up on [the discussion](https://github.com/rust-lang/docs.rs/issues/2749#issuecomment-2685521863). Maybe it also not so important to have the portal links in the index page of the docs.

README.md
```diff
@@ -142,6 +142,7 @@ Thanks to the following for helping make this library better, whether through co
 - @Sytten
 - @thewh1teagle
 - @tmpfs
+- @unkcpz
```
unkcpz, line 145: in the last commit, I add my name. It takes couple hours to figure out how to make this work ;) Fine if it is removed if such contribution is regard as too small.

Cargo.toml
```diff
@@ -77,4 +77,7 @@ rprompt = "2"
 whoami = "1.5"
 
 [package.metadata.docs.rs]
-features = ["apple-native", "windows-native", "sync-secret-service", "crypto-rust"]
+default-target = "x86_64-unknown-linux-gnu"
+targets = ["x86_64-unknown-linux-gnu", "aarch64-apple-darwin", "aarch64-apple-ios", "x86_64-pc-windows-msvc"]
+features = ["apple-native", "windows-native", "linux-native-sync-persistent", "crypto-rust"]
```
unkcpz, line 82: I change it to `linux-native-sync-persistent` since it is the one on linux can contains all `keyutils`, `dbus` and `keyutils-persistent`.

README.md
brotskydotcom: That's great you added your name. Thanks!

unkcpz:
Clearly there is a bug in CI file https://github.com/hwchen/keyring-rs/actions/runs/13550239365/workflow. 
Thanks for turn that on. I'll give it a check.

.github/workflows/ci.yaml
```diff
@@ -129,3 +129,19 @@ jobs:
...
+
+    steps:
+      - name: Fetch head
+        uses: actions/checkout@v4
+
+      - name: Install rust stable
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: nightly
+          targets: [x86_64-unknown-linux-gnu, aarch64-apple-darwin, aarch64-apple-ios, x86_64-pc-windows-msvc]
```
unkcpz, line 144: I am not sure about the syntax for setting multiple targets. Since every change and test require approve to trigger the CI, @brotskydotcom please fill free to directly add commit and change this.

src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![feature(doc_cfg)]
```
brotskydotcom, line 1: I am not comfortable with putting unstable features into this code, because it breaks stable builds (not to mention the MRSV builds). I will comment on the PR about this more fully after I finish this review.

brotskydotcom (changes requested):
I don't want to add unstable features to the code, because it breaks stable and MSRV builds.  I will add fuller discussion of this to the PR.

.github/workflows/ci.yaml
unkcpz: Nevermind, I [test in my own crate](https://github.com/unkcpz/rsdos/actions/runs/13592911412/workflow), and works now with comma split target names. See the last commit.

src/lib.rs
unkcpz: Have a quick check at tokio, this is how to make it only confined in the docrs build, https://github.com/tokio-rs/tokio/blob/20c1fdc678a7b9947b1d12d53619c90291f995d2/tokio-util/src/lib.rs#L12

```
#![cfg_attr(docsrs, feature(doc_cfg))]
```

brotskydotcom:
I've moved this back to draft status while you consider feedback. Let me know what you think.

unkcpz:
Thanks for the review!

> I've moved this back to draft status while you consider feedback. Let me know what you think.

Sure!

> they will break standard and MSRV builds unless special (and confusing) configuration is used (see https://github.com/rust-lang/rust/issues/43781 for an extensive discussion of this).

That's true, so confine it in the scope where only docrs cfg is set should do the trick. As did in the tokio-rs by having `#![cfg_attr(docsrs, feature(doc_cfg))]`.

> they won't be used in the standard docs.rs build, so they won't show up in the standard docs.

This may not true, it is mentioned in https://docs.rs/about/builds, "All crates are built in a sandbox using the nightly release of the Rust compiler. ".

brotskydotcom:
> That's true, so confine it in the scope where only docrs cfg is set should do the trick, as did in the tokio-rs by having `#![cfg_attr(docsrs, feature(doc_cfg))]`. I'll make the change if we agree on the next point.
> 

OK, I get it. And, yes, I didn't realize the standard docs build runs on nightly - thanks for pointing that out.  Please go ahead and make the change and let's try again once you've fixed the build script.

src/lib.rs
```diff
@@ -213,10 +218,14 @@ pub mod keyutils;
...
 
-#[cfg(all(
-    any(target_os = "linux", target_os = "freebsd", target_os = "openbsd"),
-    any(feature = "sync-secret-service", feature = "async-secret-service"),
+#[cfg(any(
+    all(
+        any(target_os = "linux", target_os = "freebsd", target_os = "openbsd"),
+        any(feature = "sync-secret-service", feature = "async-secret-service"),
+    ),
+    all(target_os = "linux", feature = "linux-native-sync-persistent", doc),
```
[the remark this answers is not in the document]
unkcpz: You are right!  simpler after using `#[cfg_attr(docsrs, doc(cfg(target_os = "linux")))]`.

unkcpz:
I addressed all comments, ready for another go.

brotskydotcom (changes requested):
OK, great! This looks much cleaner, thanks.

There is still one typo in the ci script, please fix that.

Please rename the file `build-xplat-docs.sh` to `test-docsrs-build.sh` since that's actually what it's for.

brotskydotcom:
Also please fix the CI script as suggested in the failed build logs.

.github/workflows/ci.yaml
```diff
@@ -129,3 +129,20 @@ jobs:
...
+    steps:
+      - name: Fetch head
+        uses: actions/checkout@v4
+
+      - name: Install rust nightly
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: nightly
+          targets: x86_64-unknown-linux-gnu, aarch64-apple-darwin, aarch64-apple-ios, x86_64-pc-windows-msvc
+          components: rust-src
```
unkcpz, line 145: Didn't realize `Cargo.lock` is not there, this should fix the failed CI.

unkcpz (commented):
Thanks for the nice review, requests addressed.

.github/workflows/ci.yaml
brotskydotcom: Yes, because this is a library, there's no Cargo.lock.

What do you think about moving the docs build out of the primary CI script and restricting it to just run on PRs and via explicit request, rather than on all pushes? There are really very few changes that can break the docs build, and it's an unnecessary overhead on most push requests.

unkcpz: Yes, I think that'll be better or even not having CI (can always run test script manually before release). Maybe move it to `publish.yml` to make sure the docs.rs will success? As you said, very few changes will break the docs build, so I am open to all options.

unkcpz:
not sure why `test_simultaneous_multiple_create_delete_single_thread` failed, I guess it is independent of changes here?

.github/workflows/ci.yaml
unkcpz: Please also check the new commit I added https://github.com/hwchen/keyring-rs/pull/237/commits/8a5c6dd838c701fd72cdf17cf4d10328d412e461
It failed in linux target because lack of libdbus dependency, but the failure not catch in bash run. I add `set -e` to make it fast failed.

brotskydotcom:
> not sure why `test_simultaneous_multiple_create_delete_single_thread` failed, I guess it is independent of changes here?

Yes, we see this from time to time due to resource starvation in the CI process. I just rerun in those cases.

.github/workflows/ci.yaml
brotskydotcom: I don't like the idea of having it in publish, because that's only run at release time. I'd rather have it be separate and run on pull requests, because I always pull to master before I release. Can you please move it to a separate workflow (called `doctest`)? Also I think if you make that a matrix workflow for all the desired platforms then you may not need a shell script? I think you might be able to just have the command line for the matrix build be what gets run in the loop in the current shell script.

unkcpz: > I'd rather have it be separate and run on pull requests, because I always pull to master before I release. Can you please move it to a separate workflow (called doctest)?

Sounds good, I like the idea :) 
Changed. 

> Also I think if you make that a matrix workflow for all the desired platforms then you may not need a shell script?

Using matrix might be bring overhead because for each target in the matrix it is an independent run. I move the bash script content into CI directly, which is more secure I think. Since otherwise changing script may trigger unexpected CI run.

brotskydotcom (changes requested):
Yes, very nice! Just one last little change and it's ready to pull, thanks!

.github/workflows/doctest.yaml
```diff
@@ -0,0 +1,40 @@
+name: Docs Build
+
+on: [ workflow_dispatch, push, pull_request ]
```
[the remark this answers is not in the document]
unkcpz: Done, cheers!

brotskydotcom (approved):

brotskydotcom:
Great, thanks for all the work on this. I'm going to pull it now, and then fix one more issue before releasing.

brotskydotcom:
@unkcpz I just released 3.6.2 with this integrated - docs look great!! Thanks again for doing this.

unkcpz:
Thank you for the review! @brotskydotcom I looking forward to contribute more! 
We use the crate in a new feature of the tool of our team through the python wrapper. I'll report to https://github.com/hwchen/keyring-rs/issues/227 when the new feature merged and released.

PR #246, opened 2025-03-15
Make the crate compile when all the features are turned off

noib3:
`cfg!` simply evaluates to a boolean, it doesn't remove any of the blocks of the if/else expression it's used in from the crate's source code.

This meant the crate wouldn't compile on linux|macos|windows if the `secret-service`|`apple-native`|`windows-native` feature was disabled.

`#[cfg(..)]` fixes this.


brotskydotcom:
Thanks so much for quickly noticing and fixing this.  As it happens I noticed it also (working on a crate that uses this as a dependency with no default features) so I am fixing it in a slightly different way and will add a test for it. Beta 2 coming in just a few minutes.

Issue #251, opened 2025-03-25
Credentials seem to persist in memory


kamou:
contrary to what have been said in #77, it seems creds are still present in memory (heap), event if the user clears everything using zeroize.

here is a test you can do to reproduce:

first create dummy password:
```
$ secret-tool store --label='dummy-password' service dummy username ak42
```

then create and build entry.rs:
```
use keyring::Entry;
use zeroize::Zeroize;

fn main() {
    let entry = Entry::new("dummy", "ak42").unwrap();
    let pass_result = entry.get_password();
    let mut password = pass_result.unwrap();
    password.zeroize();
    // wait for user input while password is still in memory
    let _ = std::io::stdin().read_line(&mut String::new());
}

```

run it with:
```
$ gdb ../target/debug/examples/entry
(gdb) run

```

in a second terminal:

```
$ cat /proc/`pidof entry`/maps | grep heap
555555765000-555555786000 rw-p 00000000 00:00 0
```

back to gdb, press, CTRL-C
and type
```
(gdb) dump binary memory out.bin 0x555555765000 0x555555786000
```

exit from gdb and finally
```
$ strings out.bin | grep supersecurepassword
supersecurepassword
supersecurepassword
```

kamou:
on a side note, I also tested zeroize, without keyring-rs, no problem observed.
here is the code I tested for that:

```
use zeroize::Zeroize;

fn main() {
    let mut password = String::new();
    let mut password_out = std::io::stdin().read_line(&mut password).unwrap();
    password.zeroize();
    password_out.zeroize();
    // wait for user input while password is still in memory
    let _ = std::io::stdin().read_line(&mut String::new());
}
```

kamou:
this might require some amount of work, trying to patch keyring-rs to use a SecureHashMap that actually does Zeroize on drop, leads me to also modify dbus-secret-service. 😭

clickonetwo:
Hi @kamou, thanks for raising this issue! I'd like to investigate on my side. Can you tell me what version of keyring-rs you are using, what platform, and which credential store? (I'm guessing you are using the default store in the v4.0.0-rc.1 on Linux, is that correct?)

kamou:
Hey @clickonetwo,
thank YOU for your work :)
Yes, this is the version I am testing against.
FYI, I believe this may be an issue with your deps as patching keyring-rs for using `secrecy` does not fix the problem (but I did a quick and dirty fix, so not sure)

[...]

PR #253, opened 2025-03-31
Move items and secrets instead of cloning (#251)

kamou:
This commit makes sure any secret is actually moved instead of cloned or copied. making sure only one version of the secret resides in memory at any point in time

### DIFF
--- a/src/secret_service.rs
+++ b/src/secret_service.rs
@@ -331,7 +331,7 @@ impl SsCredential {
             }
         }
 
-        let items: Vec<Item<'_>> = search.unlocked.into_iter().chain(search.locked).collect();
+        let items: Vec<Item> = search.unlocked.into_iter().chain(search.locked).collect();
 
         match items.len() {
             0 => return Err(ErrorCode::NoEntry),
@@ -389,7 +389,7 @@ impl SsCredential {
         require_unique: bool,
     ) -> Result<Vec<T>>
     where
-        F: Fn(&Item<'_>) -> Result<T>,
+        F: Fn(&Item) -> Result<T>,
         T: Sized,
     {
         let collection = ss.get_default_collection().map_err(decode_error)?;
--- a/src/secret_service.rs
+++ b/src/secret_service.rs
@@ -320,8 +320,8 @@ impl SsCredential {
         #[cfg(not(feature = "encrypted"))]
         let session_type = EncryptionType::Plain;
         let ss = SecretService::connect(session_type).map_err(platform_failure)?;
-        let attrs: HashMap<&str, &str> = self.search_attributes(false).into_iter().collect();
-        let search = ss.search_items(attrs).map_err(decode_error)?;
+        let attributes: HashMap<&str, &str> = self.search_attributes(false).into_iter().collect();
+        let search = ss.search_items(attributes).map_err(decode_error)?;
         let count = search.locked.len() + search.unlocked.len();
         if count == 0 {
             if let Some("default") = self.target.as_deref() {
--- a/src/secret_service.rs
+++ b/src/secret_service.rs
@@ -265,7 +265,7 @@ impl SsCredential {
     ///
     /// The created credential will have all the attributes and label
     /// of the underlying item, so you can examine them.
-    pub fn new_from_item(item: Item) -> Result<Self> {
+    pub fn new_from_item(item: &Item) -> Result<Self> {
         let attributes = item.get_attributes().map_err(decode_error)?;
         let target = attributes.get("target").cloned();
         Ok(Self {
@@ -312,7 +312,7 @@ impl SsCredential {
     /// credential for each of the matching items.
     pub fn map_matching_items<F, T>(&self, f: F, require_unique: bool) -> Result<Vec<T>>
     where
-        F: Fn(Item) -> Result<T>,
+        F: Fn(&Item) -> Result<T>,
         T: Sized,
     {
         #[cfg(feature = "encrypted")]
@@ -337,7 +337,7 @@ impl SsCredential {
             0 => return Err(ErrorCode::NoEntry),
             n if n > 1 && require_unique => {
                 let creds: Result<Vec<Box<Credential>>> = items
-                    .into_iter()
+                    .iter()
                     .map(|item| {
                         Self::new_from_item(item).map(|cred| Box::new(cred) as Box<Credential>)
                     })
@@ -349,7 +349,7 @@ impl SsCredential {
 
         let mut results: Vec<T> = Vec::with_capacity(items.len());
 
-        for item in items.into_iter() {
+        for item in items.iter() {
             item.unlock().map_err(decode_error)?;
             results.push(f(item)?);
         }
@@ -387,7 +387,7 @@ impl SsCredential {
         require_unique: bool,
     ) -> Result<Vec<T>>
     where
-        F: Fn(Item<'_>) -> Result<T>,
+        F: Fn(&Item<'_>) -> Result<T>,
         T: Sized,
     {
         let collection = ss.get_default_collection().map_err(decode_error)?;
@@ -398,7 +398,7 @@ impl SsCredential {
             0 => return Err(ErrorCode::NoEntry),
             n if n > 1 && require_unique => {
                 let creds: Result<Vec<Box<Credential>>> = items
-                    .into_iter()
+                    .iter()
                     .map(|item| {
                         Self::new_from_item(item).map(|cred| Box::new(cred) as Box<Credential>)
                     })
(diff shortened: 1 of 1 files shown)


brotskydotcom (commented):
I don't understand why we are moving items. Items never have any secret information in them, and items never refer to passwords/secrets so keeping a reference to them doesn't in any way keep a reference to a password/secret.

kamou:
ok, I can revert that part then

src/secret_service.rs
```diff
@@ -380,28 +387,33 @@ impl SsCredential {
         require_unique: bool,
     ) -> Result<Vec<T>>
     where
-        F: Fn(&Item) -> Result<T>,
+        F: Fn(&Item<'_>) -> Result<T>,
```
brotskydotcom, line 390: Please remove the unnecessary lifetime indicator on this borrow.

src/secret_service.rs
```diff
@@ -320,33 +319,41 @@ impl SsCredential {
...
         }
-        for item in search.locked.iter() {
+
+        let mut results: Vec<T> = Vec::with_capacity(items.len());
+
+        for item in items.iter() {
             item.unlock().map_err(decode_error)?;
             results.push(f(item)?);
         }
+
```
brotskydotcom, line 356: This restructuring is incorrect.  These are the designed cases:

n == 0 && require_unique: return a NoEntry error
n > 1 && require_unique: return an Ambiguous error with new items created from the attributes of the found items (without unlocking them)
! require_unique: return the result of mapping the function over every found item, but BEFORE mapping it unlock any item that needs to be unlocked. Do NOT attempt to unlock already unlocked items (both inefficient and can produce spurious side effects in terms of UI).

Given that all we are trying to do is to make sure the result of the function becomes owned by the caller, I think the original code is actually just fine, because we aren't trying to protect items, just function results. The caller's use of remove is enough to transfer the one copy to the caller.

src/secret_service.rs
```diff
@@ -380,28 +387,33 @@ impl SsCredential {
...
         }
-        let mut results: Vec<T> = vec![];
-        for item in search.iter() {
+
+        let mut results: Vec<T> = Vec::with_capacity(items.len());
+
+        for item in items.iter() {
             results.push(f(item)?);
         }
+
```
brotskydotcom, line 416: same comment as above: this restructure is wrong (and not needed). The only difference between the code here and that above is that we are doing a collection search so the results are never locked.

brotskydotcom (changes requested):
I don't think any of the changes to the body of `map_matching_items` are necessary, and most of them are incorrect (see comments below). I do like the removal of the macro in `map_matching_items`, though, please keep that. Having macros makes it very hard to know what code is doing.

To accomplish the goal of this PR, as I understand it (and what you have taught me - thanks for that!), the important change is to make sure that the memory that stores a secret returned from the secret_service are actually handed back to the caller rather than being cloned anywhere inside the keyring. That just requires changes in the edge code (get_password, get_secret) because the decode_password function is careful never to allocate even in the case of an error.

src/secret_service.rs
```diff
@@ -320,33 +319,41 @@ impl SsCredential {
...
-                    creds.push(Box::new(cred))
-                }
-                return Err(ErrorCode::Ambiguous(creds));
+
+        let attrs: HashMap<&str, &str> = self.search_attributes(false).into_iter().collect();
+        let search = ss.search_items(attrs).map_err(decode_error)?;
+
+        if search.locked.is_empty() && search.unlocked.is_empty() {
+            if let Some("default") = self.target.as_deref() {
+                return self.map_matching_legacy_items(&ss, f, require_unique);
```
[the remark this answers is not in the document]
kamou: kept it.

src/secret_service.rs
```diff
@@ -320,33 +319,41 @@ impl SsCredential {
...
-                for item in search.locked.iter().chain(search.unlocked.iter()) {
-                    let cred = Self::new_from_item(item)?;
-                    creds.push(Box::new(cred))
-                }
-                return Err(ErrorCode::Ambiguous(creds));
+
+        let attrs: HashMap<&str, &str> = self.search_attributes(false).into_iter().collect();
+        let search = ss.search_items(attrs).map_err(decode_error)?;
+
+        if search.locked.is_empty() && search.unlocked.is_empty() {
```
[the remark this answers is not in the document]
kamou: done

src/secret_service.rs
kamou: reverted

src/secret_service.rs
kamou: reverted

kamou:
question for you @brotskydotcom , do you have access to some kind of EC2 macos instance to be able to try to check whether the problem is also present on macos ?

src/secret_service.rs
```diff
@@ -320,13 +319,19 @@ impl SsCredential {
...
         let ss = SecretService::connect(session_type).map_err(platform_failure)?;
-        let attributes: HashMap<&str, &str> = self.search_attributes(false).into_iter().collect();
-        let search = ss.search_items(attributes).map_err(decode_error)?;
+
+        let attrs: HashMap<&str, &str> = self.search_attributes(false).into_iter().collect();
+        let search = ss.search_items(attrs).map_err(decode_error)?;
         let count = search.locked.len() + search.unlocked.len();
-        if count == 0 && matches!(self.target.as_ref(), Some(t) if t == "default") {
-            return self.map_matching_legacy_items(&ss, f, require_unique);
+
```
brotskydotcom, line 328: please remove

src/secret_service.rs
```diff
@@ -320,13 +319,19 @@ impl SsCredential {
...
+        if count == 0 {
+            if let Some("default") = self.target.as_deref() {
+                return self.map_matching_legacy_items(&ss, f, require_unique);
+            }
         }
+
```
brotskydotcom, line 334: please remove

src/secret_service.rs
```diff
@@ -339,6 +344,7 @@ impl SsCredential {
                 return Err(ErrorCode::Ambiguous(creds));
             }
         }
+
```
brotskydotcom, line 347: please remove

brotskydotcom (changes requested):
I'm not a big believer in blank lines in code, and introducing them makes the change look more complex than it is. Please take out the marked blank lines you introduced, thanks.

brotskydotcom:
@kamou Will you have a chance to take care of these last few changes so I can merge?

kamou:
yesss

src/secret_service.rs
```diff
@@ -321,11 +320,13 @@ impl SsCredential {
         #[cfg(not(feature = "encrypted"))]
         let session_type = EncryptionType::Plain;
         let ss = SecretService::connect(session_type).map_err(platform_failure)?;
-        let attributes: HashMap<&str, &str> = self.search_attributes(false).into_iter().collect();
-        let search = ss.search_items(attributes).map_err(decode_error)?;
+        let attrs: HashMap<&str, &str> = self.search_attributes(false).into_iter().collect();
+        let search = ss.search_items(attrs).map_err(decode_error)?;
```
brotskydotcom, line 324: Please revert this rename.

brotskydotcom (changes requested):
I am TRULY sorry to be this picky, but I'm trying to get this PR to a place where it doesn't change any lines of code that aren't relevant to what it's trying to do. Can you please revert the rename of attributes to attrs marked below? Thanks!

brotskydotcom:
> question for you @brotskydotcom , do you have access to some kind of EC2 macos instance to be able to try to check whether the problem is also present on macos ?

I've just run the test on macOS. The problem doesn't exist there. I'm still going to make one safety change in the way that clone is done (using to_owned rather than to_vec, because to_owned is what's recommended for the keychain data object that has the data), but it looks like the MacOS keychain dealloc routines clear the memory (or return it to the kernel, it's hard to say which).

brotskydotcom (approved):
Thanks for all your work on this. I'm going to merge it.

kamou:
@brotskydotcom  now we need tests to make sure that any new contributions will not bring back passwords to memory.

brotskydotcom:
Do you have any suggestions for how to test this? I guess we could have a test program like yours and shell scripts for each platform. I can't really think of any other approach.

kamou:
Yup, for keyring-rs, my test can be easily scripted. 
For dbus-secret-service I'll need to think about a way to log the shared key without buffering it.
That may require unsafe code (to call libc's write function) for debug mode only. But I'm not 100% sure yet.

kamou:
something like this:
https://github.com/kamou/nextcloud-rs/blob/main/src/bin/common/mod.rs#L97

but I'll see if we can avoid unsafe.

Issue #232, opened 2024-12-24
Clean up selection of credential store


brotskydotcom:
(Requested by @soywod)

The current usage pattern for this crate is that the developer chooses the credential store to use (for each platform) at compile time, and the users of the developer's application have no choice in the matter. Because it's not possible, currently, to build this crate with all possible credential stores on all platforms, there is no way for a developer to let his application users choose their desired credential store at runtime.

Perhaps, in the next major version of keyring, this could change?

brotskydotcom:
I'm not sure this requires any enhancement. The only two credential stores you can't use at the same time are the synchronous and asynchronous versions of the secret service, and those are the same credential store, so there would be no functional difference in choosing one or the other.  In all other cases, you can build in all the credential stores and then choose between them at runtime by using `set_default_credential_builder`. @soywod, can you say more about what flexibility you are trying to get?

soywod:
> so there would be no functional difference in choosing one or the other.

This may be partially wrong. So far the `dbus` crate exposes a blocking API (std) and a nonblock API (tokio). The `zbus` crate exposes by default an async API (async-io), another async API behind a cargo feature (tokio) and a blocking API. But you cannot use the blocking API only, so it comes with `futures-*` dependencies. Both do not manage the same way: one needs a third party lib (`libdbus`), the other spawns an async execute (although looks like you can [disable it](https://github.com/dbus2/zbus/tree/main/zbus#compatibility-with-async-runtimes)).

My point is: some people may need the blocking D-Bus API, some the Z-Bus tokio API, some both. If you expose everything, people can compose the way they want. This idea brings us to the last point (and joins your [questions](https://github.com/hwchen/secret-service-rs/issues/82#issuecomment-2307351843)): as a lib you don't want to rewrite from scratch every implementations you have for std, tokio, async-std etc. Code needs to be put in common.

---

To summarize: ideally, a reusable and composable lib should:

- Have cumulative cargo features and not exclusive ones
- Expose components both at compile time (cargo features) and runtime
- Not care about I/O and runtime, no need to deal with duplicate async/non async traits
- Propose default I/O connectors, and make sure lib consumers can develop their own connector if needed

[...]

PR #257, opened 2025-04-22
Merged cargo features inside one unique platform-store

soywod:
Reference : #232

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -121,7 +121,7 @@ jobs:
           target: aarch64-apple-ios
 
       - name: Build iOS library
-        run: cargo build --target aarch64-apple-ios --features=apple-native --example=iostest
+        run: cargo build --target aarch64-apple-ios --example=iostest
 
   msrv_native:
     runs-on: ubuntu-latest
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,14 +13,10 @@ exclude = [".github/"]
 readme = "README.md"
 
 [features]
-default = ["apple-native", "secret-service", "windows-native"]
+default = ["platform-store"]
 
-## Use the built-in Keychain Services on macOS and iOS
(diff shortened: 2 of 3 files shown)


brotskydotcom:
This is not a useful simplification, because combining all the platform dependencies into a single feature doesn't allow someone to use, for example, the apple-native support on macOS while using linux-keyutils (but NOT secret-service) on linux (because clients can't enable dependencies unless there's a specific feature they can use to enable them). So I'm going to close this pull request.

What I was hoping you were going to do was to create a separate crate that does what your "linux-keyutils-persistent" keystore used to do, that is, combine the functionality of the linux-keyutils keystore and the secret-service keystore. The Cargo.toml for that crate would then be a great example of how you build using multiple keystores on the same platform.

soywod:
> What I was hoping you were going to do was to create a separate crate that does what your "linux-keyutils-persistent" keystore used to do, that is, combine the functionality of the linux-keyutils keystore and the secret-service keystore. The Cargo.toml for that crate would then be a great example of how you build using multiple keystores on the same platform.

I see. So you confirm that everything we said in the other thread was done except for this linux-keyutils-persistent?

brotskydotcom:
>  So you confirm that everything we said in the other thread was done except for this linux-keyutils-persistent?

Yes, exactly. I’ve submitted PRs for the crate extensions for the old features sync-secret-service (keystore feature in dbus-secret-service), async-secret-service (keystore feature in secret-service), and linux-keyutils (keystore feature in linux-keyutils), and I am working with the maintainers to get them accepted. The old sync-secret-service feature (now called just secret-service) will remain the default in this crate (and use the keystore feature in dbus-secret-service). So what we need now is a new crate that combines them into a single keystore.

Feel free to wait until I can get all the other crates released with the PRs in them 🙂. I have to wait for that before I can release v4.0.0 of keyring, as well.

PR #261, opened 2025-07-02
minor cleanup of unsafe pointer ops in windows.rs

oconnor663:
I don't think there were any bugs here, just a chance to tighten up some of the pointer manipulation. Please let me know what you think.

### DIFF
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -41,7 +41,6 @@ use super::error::{Error as ErrorCode, Result};
 use byteorder::{ByteOrder, LittleEndian};
 use std::collections::HashMap;
 use std::iter::once;
-use std::mem::MaybeUninit;
 use std::str;
 use windows_sys::Win32::Foundation::{
     ERROR_BAD_USERNAME, ERROR_INVALID_FLAGS, ERROR_INVALID_PARAMETER, ERROR_NO_SUCH_LOGON_SESSION,
@@ -251,7 +250,7 @@ impl WinCredential {
         };
         let attribute_count = 0;
         let attributes: *mut CREDENTIAL_ATTRIBUTEW = std::ptr::null_mut();
-        let mut credential = CREDENTIALW {
+        let credential = CREDENTIALW {
             Flags: flags,
             Type: cred_type,
             TargetName: target_name.as_mut_ptr(),
@@ -265,10 +264,8 @@ impl WinCredential {
             TargetAlias: target_alias.as_mut_ptr(),
             UserName: username.as_mut_ptr(),
         };
-        // raw pointer to credential, is coerced from &mut
-        let p_credential: *const CREDENTIALW = &mut credential;
         // Call windows API
-        let result = match unsafe { CredWriteW(p_credential, 0) } {
+        let result = match unsafe { CredWriteW(&credential, 0) } {
             0 => Err(decode_error()),
             _ => Ok(()),
         };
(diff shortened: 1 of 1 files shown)


src/windows.rs
```diff
@@ -289,37 +286,25 @@ impl WinCredential {
...
                 // `CredReadW` failed, so no allocation has been done, so no free needs to be done
                 Err(decode_error())
             }
             _ => {
-                // `CredReadW` succeeded, so p_credential points at an allocated credential.
-                // To do anything with it, we need to cast it to the right type.  That takes two steps:
-                // first we remove the "uninitialized" guard around it, then we reinterpret it as a
-                // pointer to the right structure type.
-                let p_credential = unsafe { p_credential.assume_init() };
-                let w_credential: CREDENTIALW = unsafe { *p_credential };
```
oconnor663, line 318: This line was making a bitwise copy of the `CREDENTIALW` struct from the heap to the stack, which might not've been intended?

clickonetwo:
Thanks for this! I didn't write this code in the first place; Walther did. It was before I came on board. The comments are all mine doing code archaeology but I never really felt comfortable enough with the Windows system calls to touch it much.

This certainly looks a lot cleaner to me, but I would love to get a review from someone who does a lot of Windows-level system coding in Rust. Are you such a person? Do you know any?

I've approved the workflows so it will get some testing, but have you done testing of your own?

src/windows.rs
```diff
@@ -289,37 +286,25 @@ impl WinCredential {
...
-                // Now we can apply the passed extractor function to the credential.
-                let result = f(&w_credential);
+                // `CredReadW` succeeded, so p_credential points at an allocated credential. Apply
+                // the passed extractor function to it.
+                let result = f(unsafe { &*p_credential });
```
clickonetwo, line 305: The combination of the implicit coercion and unsafe as part of the function call makes this code hard to read.  What do you think about this instead:
```
let ref_cred: &mut WCREDENTIAL = unsafe { &*p_credential };
let result = f(ref_cred)
```
I'm willing to be talked out of it, but I like the clarity of separating the two steps.

src/windows.rs
```diff
@@ -289,37 +286,25 @@ impl WinCredential {
...
                 // Finally, we erase the secret and free the allocated credential.
-                erase_secret(&w_credential);
+                erase_secret(unsafe { &mut *p_credential });
```
clickonetwo, line 307: The change above makes this:
```
erase_secret(ref_cred)
```
With both these changes, I think it's easy to understand. Applying & one place and &mut feels incredibly unsafe, and I'd rather just make it clear that it's a mutable block of memory that we are passing as non-mut and then as mut.

clickonetwo (changes requested):
This looks generally good.  I have one hygiene suggestion, since this is meant to be a hygience PR.

src/windows.rs
```diff
@@ -445,7 +430,7 @@ fn extract_secret(credential: &CREDENTIALW) -> Result<Vec<u8>> {
     Ok(blob.to_vec())
 }
 
-fn erase_secret(credential: &CREDENTIALW) {
+fn erase_secret(credential: &mut CREDENTIALW) {
```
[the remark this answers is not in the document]
clickonetwo: In this specific case, I think it's enough that you made it clear the credential is being modified with the &mut. That's really what we need to make sure Rust knows about, so it can ensure that no one else is mucking with this credential at the same time.

In general, I think confining unsafe to the code that is actually unsafe is the better plan. I don't believe that propagating unsafe upwards as a way of signaling that unsafe code is inside is helpful. Better to document that the code makes system calls so there is no guarantee that a bug won't crash things.

PR #265, opened 2025-07-26
windows: use static value for 'comment' attribute

### LINKED ISSUE #264: Can't seem to create a credential on Windows (with a generated target)
First off, thanks for this crate :-) All in all, it's been a wonderful abstraction layer to have available in an otherwise complex area.

I'm getting the following error when I create an entry on Windows, but only on a virtual desktop (VDI). I am not able to reproduce on my Windows 11 bare metal.

```rust
keyring::Entry::new(
  &"https://xxxxxxxxx.xxxx.com/xxxxxxxxxxxxxxxxxxxxxxxxx",
  &"xxxx.xxxxxxx.xxxxxx.a9e6eb1a-7ec2-4ffc-bba5-dcf9a5b311e9.Authorization",
)
// keyring::Error::TooLong("comment", 256)
```

All told, there are 122 characters here. Well under the advertised platform maximum of 256 characters for the `comment` field.

https://github.com/open-source-cooperative/keyring-rs/blob/9635a2f53a19eb7f188cdc4e38982dcb19caee00/src/windows.rs#L383 is, I believe, where the error originates – specifically with `format!("{user}@{service}:{user}.{service} (keyring v{VERSION})")`. 

It's my understanding from the comments in that function (below) that the `comment` attribute is unimportant (attribution notwithstanding 🙂). In theory, I would be able to customize it with `Entry::update_attributes()`, but the error is bring tripped on `Entry::new()`, so there's no opportunity to call that API.

https://github.com/open-source-cooperative/keyring-rs/blob/9635a2f53a19eb7f188cdc4e38982dcb19caee00/src/windows.rs#L345-L386

## Investigation notes

This check dates back to 130b6fa (Add linux and windows implementations., 2022-10-01). It was merged in a24df2b (Merge pull request #98 from brotskydotcom/traits, 2022-10-08). It doesn't appear that other backends validate attributes at this stage. Is this actually necessary to do? The merged commits don't bring attention to this line. I suspected maybe this *used* to be done on all clients and Windows was missed in some later update, but this doesn't appear to be the case.

1. Is this check still necessary?
2. Assuming it is, is a verbose comment necessary to add by default? (It can be made verbose again with `Entry::update_attributes()`, if desired.)

vermiculus:
windows: use static value for 'comment' attribute

Some Windows environments have a severely restricted 256 characters
allowed for a 'comment'. When the 'user' and 'service' attributes are
lengthy (e.g., client IDs and URLs), this information can exceed the
limit -- particularly in the 'target == None' case (where both 'user'
and 'service' are duplicated).

To add to the trouble, the call to `credential.validate_attributes()`
before this function returns prevents a consumer from updating this
comment to be within the length limit.

Switch to a static string for this comment that is within any
reasonable length limit. If a consumer would like to add their own
more verbose comment, they can do so with `Entry::update_attributes()`
after the entry is created.

Note: a pass-through optional parameter for the comment value was
considered, but there's not really a good, obvious answer for a clean
API on Entry with such an approach. The 'comment' attribute also
appears to be Windows-specific. This approach should not have any
impact on consumers and certainly does not have an impact on the API.

An alternate-alternate approach may be possible with just removing the
call to `credential.validate_attributes()`, but I don't have the
expertise to understand what the effects of this would be.

---

Fixes #264.

### DIFF
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -362,7 +362,7 @@ impl WinCredential {
                 username: user.to_string(),
                 target_name: target.to_string(),
                 target_alias: String::new(),
-                comment: format!("{user}@{service}:{target} (keyring v{VERSION})"),
+                comment: format!("keyring v{VERSION}"),
             }
         } else {
             Self {
@@ -377,7 +377,7 @@ impl WinCredential {
                 username: user.to_string(),
                 target_name: format!("{user}.{service}"),
                 target_alias: String::new(),
-                comment: format!("{user}@{service}:{user}.{service} (keyring v{VERSION})"),
+                comment: format!("keyring v{VERSION}"),
             }
         };
         credential.validate_attributes(None, None)?;


brotskydotcom (approved):
This fix seems quite reasonable. Probably even more robust to have no comment at all, and not do any validation of the comment, because the comment is just a decoration, but this is a nicely minimal fix and at least notes that keyring last wrote the entry.

src/windows.rs
```diff
@@ -362,7 +362,7 @@ impl WinCredential {
                 username: user.to_string(),
                 target_name: target.to_string(),
                 target_alias: String::new(),
-                comment: format!("{user}@{service}:{target} (keyring v{VERSION})"),
+                comment: "keyring v{VERSION}".to_string(),
```
brotskydotcom, line 365: Looks like you need the format! statement to convert the {VERSION}.

vermiculus: Ah! Another lesson for me to review my patches with a full night's rest :-) I've pushed a fix.

brotskydotcom:
Thanks for this! Also please see my comment in #264 and get back to me there.

PR #302, opened 2026-01-21
add db-keystore with name "sqlite"

stevelr:
You asked for PRs with new keystore implementations, so here's one!

New keystore backend with sqlite storage (implemented with turso), with
- optional encryption
- ambiguous or non-ambiguous (service,user) pairs
- search with filters on service, user, uuid, and comment

See [db-keystore](https://github.com/stevelr/db-keystore) README for more info

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ exclude = [".github", "python", "tauri"]
 [dependencies]
 base64 = "0.22.1"
 clap = { version = "4.5.54", features = ["derive"] }
-db-keystore = "0.3.0"
+db-keystore = "0.3.1"
 fastrand = "2.3.0"
 keyring-core = { version = "0.7.2", features = ["sample"] }
 rpassword = "7.4.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -680,9 +680,9 @@ dependencies = [
 
 [[package]]
 name = "db-keystore"
-version = "0.3.0"
+version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0bddf2d511195c9d83644518a60ef928085554ae41f4b412dd5f28c4900120a9"
+checksum = "98e7a76c30fd9fcb0b9e46a848028bb82d8cc9854fb3d4f9db8851b7db2648fe"
 dependencies = [
  "futures",
  "keyring-core",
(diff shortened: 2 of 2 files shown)


brotskydotcom (changes requested):
This crate doesn't use features so that people can use it to try out every available keystore. So you can't remove the optional qualifier on your crate and remove the features.

brotskydotcom:
Can you please sign your commits? This crate requires all commits to be signed.

brotskydotcom:
Thanks a lot for submitting this!! It will be great to have another cross-platform keystore available.

Before I can accept this PR, however, I noticed that there are a couple of issues with the keystore itself that I am hoping you can address:

1. (major) The keystore contract is that secrets can contain arbitrary binary data. I notice [here](https://github.com/stevelr/db-keystore/blob/3b007214ab9fe29c0cb18f3043e9d115aa8d0e38/src/lib.rs#L946C5-L947C95) that set-secret will fail on non-UTF8 data. Can you instead transparently tag and encode that data (e.g., using base64 without padding) when storing and then decode and untag it when reading?
2. (major) Can you please provide an example program that shows how to create and use an encrypted database (including creating an appropriate key), so the crate stands alone even without this PR?
3. (minor) Your doc says that the default path is at `~/.local/state/` but your [default path function](https://github.com/stevelr/db-keystore/blob/3b007214ab9fe29c0cb18f3043e9d115aa8d0e38/src/lib.rs#L113) uses `~/.local/.state`.
4. (minor) Your code seems to fail if you use the default path but it doesn't yet exist. I suggest that you create the `~/.local/state` directory if necessary when the default path function is invoked.
5. (minor) Can you clarify in your docs whether bit's the entire database file that's encrypted, whether it's both keys and data that are encrypted or whether it's just the data that's encrypted? Also what the available algorithms for encryption are? Also what happens if you supply a different encryption key when opening an existing encrypted database than the one that was used to create it? (For example, is this noticed by the code, or do you just get garbage data?) Also whether the effectiveness of the encryption depends on the key?

Cargo.toml
```diff
@@ -8,13 +8,14 @@ keywords = ["password", "credential", "keychain", "keyring", "cross-platform"]
...
 version = "4.0.0-beta.3"
-rust-version = "1.88.0"     # uses let chains
+rust-version = "1.88.0"                                                         # uses let chains
 edition = "2024"
 exclude = [".github", "python", "tauri"]
 
 [dependencies]
 base64 = "0.22.1"
 clap = { version = "4.5.54", features = ["derive"] }
+db-keystore = "0.3.0"
```
brotskydotcom: Thanks for sorting this, too!

brotskydotcom (approved):
This looks great, thanks! And thanks for signing your commits!

I'll pull this as soon as you've had a chance to look at my review comments on the underlying keystore, which I'll post now in an issue there. I really appreciate all your work here, and look forward to merging!

PR #312, opened 2026-04-22
Prepare release of v4 final

brotskydotcom:
Now that keyring-core is at v1.0.0, and all the included credential stores use that version, it's time to release v4.0.0 of this crate.


brotskydotcom:
This can't be published yet, because db-keystore v0.4.2-pre.2 hasn't been released to crates.io. But the git dependency should allow the pull request to build properly.

Issue #322, opened 2026-06-08
Move all keystore implementations to features


eirnym:
Could you please move all keystore implementations to features? all of them could stay as defaults.

I understand feature defaults, but I'd like to enable all features as I use them. This includes very deep and wide list of dependenceis I'd like to exclude from my application.

brotskydotcom:
Hi @eirnym, thanks for this report.

This is simply a sample code vault. That's why it loads all the keystores and other dependencies: to provide samples of how you use all of them.

If you are writing an app that uses the keyring then you don't need this crate at all: use the [keyring-core](https://crates.io/crates/keyring-core) crate and link and exactly those keystores you want to use. This is explained more fully in [these docs](https://github.com/open-source-cooperative/keyring-rs/wiki/Keyring).

Also take a look at #317 and see if that's what you're hoping for.

Let me know if this answers your question, thanks.

eirnym:
But it isn't documented here: https://docs.rs/keyring/latest/keyring/

brotskydotcom:
Great point, it's in the [crate README](https://crates.io/crates/keyring) but not in the library docs. Thanks, I'll make this a doc bug and fix it.

eirnym:
In rust ecosystem for many crates you don't want to import `-core` modules, because they're internal. Also it worth to admit, that implementation of `keyring` crate is good to be reusable, and not copy-pasted anywhere else, when I need to use some default per-platform or give a user an option to choose which keyring store to use.

Thus putting all of stores feature-gated will solve both issues.

If you'd truly want to have a keyring example, why not to rename crate?

brotskydotcom:
There is a long, long history that answers your questions; you can read about it in the closed issues. I really think you are asking for the same thing as #317: to make this crate usable as a standalone (the way it used to be). You're just asking for the client to pick the credential stores by using features rather than the crate picking them. We tried that in v3 and it was not popular. I'll see what I can do.

In the meantime, just use keyring-core and the credential stores directly. There's really no reason not to, your concern about naming notwithstanding.

eirnym:
@brotskydotcom I published an idea preview how it could be implemented. Please assess general direction and if I'm allowed to continue.

brotskydotcom:
Thanks very much, @eirnym, for submitting #323. For the reasons I explain in that PR, I don't want to go in that direction.

I am still leaving this bug open as a doc bug, and I will fix it in my fix for #317. Thanks again for pointing out the deficiency in the docs.

[...]

Issue #317, opened 2026-05-03
feat: default credential store that uses the appropriate store based on the platform/os


ossenthusiast:
This would be extremely helpful for application developers who just wanna add support for cross-platform keyring support to their application. WDYT?

brotskydotcom:
Hi there, thanks so much for this suggestion. To understand why things don't work this way, I need to review a little history.

* The original [keyring crate](https://crates.io/crates/keyring) worked the way you are suggesting. It only had one credential store for each platform, so it automatically picked whichever one was present on that platform.
* As time went on, and multiple stores became available for some platforms, the keyring crate added features to allow selecting the default store on each platform. This turned out to be both a confusing and a limiting way to go, because some applications actually want to use multiple stores on a single platform.
* Based on client feedback about multiple stores, default stores, and how to specify different options, I decided to actually split off the implementation of the keyring core API from the implementation of the various credential store. At that point, the API crate (this one, keyring-core) stopped having any knowledge of which credential stores are available and which have been linked with any given application and so there's no way for it to have the capability you're suggesting.

But! Even though what you're asking for is not possible for keyring-core, that doesn't mean it's not a useful idea or that it can't be pre-supplied 😊. The original keyring crate has been migrated to be a source of sample code for use of the keyring-core and credential stores, and in that crate you will find [the use-native-store function](https://github.com/open-source-cooperative/keyring-rs/blob/3a46e9ce493aa0b63be84e688cc0723acf6714c5/src/lib.rs#L74-L107) which does pretty much exactly what you're asking for. My suggestion is that you add that function (or something similar) to the source of your application.

If this addresses your concern sufficiently, feel free to close this issue. If not, by all means reply here and I'll take a look.

ossenthusiast:
❤️ Thank you for the detailed response! Looks like `use_native_store` is exactly what I was looking for.

necauqua:
Hi, I was looking in frustration for a long while exactly for this.

Story:
Building a small toy side-project that would need to store a credential for some server api.
Want to just use a sane reasonable default while keeping it crossplatform (likely only going to be used on linux/windows).
Surprised to see `keyring v4.0.0`, a major (and recent, cuz .0.0) semver update, open the readme - "we reworked everything and you DO NOT WANT TO (big bold letters, could've been red) depend on this crate, use keyring-core".
Okay, looking into keyring-core - "we dont provide any store impls other than test ones".

[...]

Issue #259, opened 2025-05-28
v4: separate core library from credential store implementations


brotskydotcom:
I am about to split this crate into several crates:

- keyring-core (the cross platform library parts of this crate)
- apple-native-keyring (the apple-native credential store implementations)
- windows-native-keyring (the windows-native credential store implementation)
- linux-keyutils-keyring (the linux-keyutils credential store implementation)
- dbus-secret-service-keyring (the sync secret-service credential store implementation)
- zbus-secret-service-keyring (the async secret-service credential store implementation)
- keyring (the CLI part of this crate done as an installable main app, plus re-export of the core library with default stores)

That will make it a lot easier to add evolve each of the credential stores separately, and for users to add new keystores independently. It will also allow this crate to have features which control exactly which credential stores get built, and to give examples of how to do this for all the different credential store crates.

The v4 release of this crate will be done once the split is complete.

catgirlconspiracy:
Ok, so this is the actual plan now vs what's in the last rc and what's described in #232 ? I saw that yesterday and was kinda worried because for me the tradeoffs of forcing libdbus are Bad Actually. I already have zbus and tokio, and a C dependency, even vendored, is an additional complication when crosscompiling to other Linux architectures with the -musl targets. If your entire build is Rust, all you need is the target stuff from rustup, but once you have C you have to provide a C compiler (well, you can just use clang because GCC cross toolchains are miserable) as well as libc headers for your target.

Splitting the crate would be fine though. I assume the "default stores" in the keyring crate would use libdbus?

And would it make sense to yank the v4 prerelease versions, at least the rc? I was tempted to update to it to get ahead of the changes, but that would've been wasted effort if the plan has changed so drastically since that rc.

brotskydotcom:
Thanks for this comment! It feels to me like confirmation that clients of keyring have lots of different needs and would welcome having complete control over their choice of credential store modules (rather than relying on defaults).

Yes, I have been thinking about yanking the published rc builds because they are, indeed, very misleading. I'm glad to hear an voice outside my head say the same thing :). I have now yanked all the 4.0 builds.

brotskydotcom:
[...]

Issue #97, opened 2022-10-03
Version 2: plug-in credential providers, and mocking


brotskydotcom:
OK, it's time for a v2 that addresses the issues raised in #94 and #95 (and provides the ability for clients to address the issues in #76 and #84).  Here are the enhancements needed:

1. The ability for 3rd parties to provide new credential storage mechanisms, and for clients to choose those as their default mechanism.
2.  Support for a "mock" credential store that allows controlling both return and error values.  Preferably the mock store should also be an example of how you plug in a new storage mechanism.

These enhancements should not complicate the existing (very simple) `Entry` model, so that existing clients don't have to revise their code.

While we're at it, we should clean up the API so that creating entries can fail.  This will be a breaking change, but one that existing clients can adapt to simply by adding `expect` calls to their existing entry creations.

brotskydotcom:
Here are some design notes:

1. Platform credential stores now work by supporting two traits: the `CredentialAPI` (for operating on entries) and the `CredentialBuilderAPI` (for creating new entries).
2. The platform-independent `Error` class has been extended to handle parameter errors other than just "too long".
3. All the new traits are Sync and Send so that the interface is both thread-compatible and thread-safe.
4. The default platform providers have been rebuilt as pluggable credential stores.
5. Entries use default platform providers unless a different default provider is requested (and provided) by the client.
6. The `Entry::new_with_credential` API allows specifying a 3rd part credential for the entry.

Docs are still in process.

landhb:
One note I noticed on the interface. Instead of having a mutable global that requires users to call:

```rust
// set the proper backend
keyring::set_default_credential_builder(Box::new(LinuxNativeCredentialBuilder));

// start an entry
let entry = Entry::new("myservice", "keyid-2").unwrap();
```

Could we instead simply add another `Entry` method, which users could use like:

```rust
Entry::new_with_builder(&LinuxNativeCredentialBuilder, "myservice" , "keyid-2")?;
```

The implementation could simply be:

```rust
/// Create an entry for the given target, service, and username.
/// The default credential builder is used.
pub fn new_with_builder(builder: &CredentialBuilder, service: &str, user: &str) -> Result<Entry> {
    let credential = builder.build(None, service, user)?;
    Ok(Entry { inner: credential })
}
```

That then removes the need for the mutable global and makes the call thread safe.

landhb:
For the mock usage, users could do something like:

```rust
#[cfg(test)]
[...]

PR #323, opened 2026-06-15
Pluggable keyring preview

eirnym:
It's an idea preview for #322 how to make keyring pluggable and don't provide false hopes, that some engine is available

done in current idea preview:
* exported bin as a separate sub-crate to remove not strictly required dependencies and features.
* loosen restrictions of `dbus` & `zbus` availability (e.g zbus is available even on Windows and dbus is available via `MacPorts` & `Homebrew`).
* added a documentation line about `APP_SANDBOX_CONTAINER_ID` env variable.
* Moved `turso` and a `sample` key stores to bin folder due to few issues and a lot of dependencies.

wasn't a goal (at the moment of creating PR):
* make it final product, so don't expect it's even compiles.
* provide full interface.

Possible expansion:
* Ability for a user to register crates by any name. I'd limit names to `&'static str`.
* Make `NAMED_STORES` an endpoint returning `Vec<&'static str>` to provide currently registered endpoints.
* register stores by name. This include registering `default` stores as well. 
* My proposition: to register a store `default` (or `native`) to call a default store. the dependency of dbus/zbus keystores is about adding a default keystore if available. Decision which exact store will become "native default" if few is available is up to maintainers. I can express my personal preferences if asked.

I expect:
* Comments about direction of this idea in general.


brotskydotcom:
Hi, thanks for submitting this draft, it gives me a good idea of where you are trying to go.

Keyring has a long history and everything you are doing here, including using features to enable specific credential stores, was tried in v3 and there was a lot of community feedback that it's not a good way to go. Instead, the direction taken in v4 - based on community feedback, please read the closed issues if you want to know the detailed history - was to separate out keyring-core (containing the API) and the credential stores (implementing the API) so that application authors could decide exactly which credential stores they wanted to use on which platforms. (That's why the credential stores themselves do not have platform dependencies.)

The main negative feedback on v4 (see #317) has been that there is no longer a single, standalone crate that someone can use if they don't care about which credential store is in use and just want to be able to store and retrieve secrets easily from their app. I believe I have a fix for that issue, and will be publishing it sometime in the next few weeks.

Long story short: I am not going to go in the direction suggested by this PR, but I appreciate greatly your taking the time to clarify your ideas.

eirnym:
Could you please provide me few links to some of negative feedback mentioned so I can educate myself?

brotskydotcom:
Best to review the closed issues. #232 and #259 and #97 (for v2) are good places to start.

eirnym:
Thank you for a good read. I see architectural problems with the keyring-rs as a whole. I'll give it a think

brotskydotcom:
Also wait for me to check in my current work on #317, because I think that addresses a lot of the issues you've raised.

eirnym:
Why zbus is enabled only on linux & freebsd while it's technically available on all systems (including Windows). My point is to use other selector like `cfg(all(unix, not(any(target_os = "macOS", target_os="ios", target_os="android"))))`. The rest of #325 looks as I would actually need.

brotskydotcom:
On Windows we use the Windows Credential Manager for secure storage. (Keyring has always been opinionated about that.) As the docs say: if you want to use different credential stores on a given platform, then don't link to this crate at all. Just link to keyring-core and the credential manager(s) you want to use, and set the default credential store to whatever you want on those platforms.

eirnym:
I didn't question what credential manager is used on Windows. Question was about other BSDs where change is matter of slight adjustment of `cfg` expression.

brotskydotcom2:
OK, sorry, I was misled by the Windows reference :). To be sure I understand what you're asking about/for, let me ask you which of these two (1 or 2) best describe your problem statement:

1. I am trying to use the `v1` module on a *nix other than Linux and FreeBSD but I am prevented by the current platform gate.
2. I am trying to use the `cli` module on a *nix other than Linux and FreeBSD but I am prevented by the current platform gate.

If your answer is (1), then by all means propose a new PR fixing the platform gate to work for whichever other *nix are missing. I'm not enough of a *nix expert to know how to do that, although I suspect you gave me the answer above (`cfg(all(unix, not(any(target_os = "macOS", target_os="ios", target_os="android"))))`)

If your answer is (2), then you shouldn't be doing that :). Only CLI-style or demo apps should be using the `cli` module. All other keyring-based apps should just be linking with keyring-core and credential manager crates directly.

Looking forward to hearing from you.

eirnym:
Thank you for pointing me to create a PR #328. I indeed care the most about the library part.

PR #333, opened 2026-06-26
Use `AtomicBool` to guard initialization and return error if occur on the first call.

### LINKED ISSUE #332: Log or return failed initialization
Currently if any `keyring` store implementation initialization fail, error is always hidden and never logged in any way. 

Could you please either report an error back to a user changing mechanism of "once" to something like `AtomicBool` (preferable) or log it using `log` or `tracing`.

I'd used `AtomicBool` instead of `Once` to synchronization.

eirnym:
Subsequent calls on initialization fail won't get the same result, but `Error::NoDefaultStore` instead.

Initialization is NOOP on non-windows, non-unix, iOS and Android

closes #332

### DIFF
--- a/src/v1.rs
+++ b/src/v1.rs
@@ -22,7 +22,7 @@
 //! it in the [Keyring
 //! wiki](https://github.com/open-source-cooperative/keyring-rs/wiki/Keyring).
 
-use std::sync::Once;
+use core::sync::atomic::{AtomicBool, Ordering};
 
 pub use keyring_core::{Error, Result};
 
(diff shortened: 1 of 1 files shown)


src/v1.rs
```diff
@@ -86,28 +90,19 @@ impl Entry {
...
         not(any(target_os = "macos", target_os = "ios", target_os = "android"))
     ))]
-    {
-        if let Ok(store) = zbus_secret_service_keyring_store::Store::new() {
-            keyring_core::set_default_store(store);
-        }
-    }
+    let store = zbus_secret_service_keyring_store::Store::new()?;
+    #[cfg(any(unix, windows))]
+    keyring_core::set_default_store(store);
```
brotskydotcom, line 106: This doesn't compile on iOS or android (`store` is not defined). It needs to compile on all platforms.

brotskydotcom (changes requested):
I like this direction a lot. Please fix it so it compiles on all platforms.

src/v1.rs
eirnym: thanks, I thought you included iOS & Android. Fixed and commit message updated

brotskydotcom: Do you mean include those platforms in the CI? Good point, I probably should.

brotskydotcom (approved):
Thanks for this!

